PR #16, opened 2021-10-01
WIP - Tests/db methods

machikoyasuda:


tests/conftest.py
```diff
@@ -1,13 +1,20 @@
 import pytest
 
 from eligibility_server.app import app as server
+from eligibility_server.app import Database as Database
 
 
 @pytest.fixture
 def app():
```
thekaveman, line 8: I can only see this fixture used below in the parameter list for `client()` fixture, is that right?

This is all a bit awkward with `import app as server` and then an `app()` fixture. Is it possible to use `server` directly in the `client` fixture below, e.g. 

```python
@pytest.fixture
def client():
    return server.test_client()
```

If so, the `app()` fixture can go away, and the alias in `from eligibility_server.app import app` is no longer needed. So `client()` becomes:

```python
@pytest.fixture
def client():
    return app.test_client()
```

Almost just like it is now!

tests/test_app.py
```diff
@@ -1,4 +1,9 @@
-def test_healthcheck(app, client):
-    response = client.get("/healthcheck")
+"""
+Test API
```
thekaveman, line 2: This isn't testing the API so much as the server/app itself (with a healthcheck anyway).

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,28 @@
+name: Run pytests
+
+on:
+  pull_request:
+    branches: [ "*" ]
+    paths:
+      - 'data/**'
+      - 'eligibility_server/**'
+      - 'keys/**'
+      - 'static/**'
```
thekaveman, line 10: We'll get rid of the `static/` folder in this repo soon

thekaveman (changes requested):
Really great test coverage for the database. Couple notes on the fixtures.

Issue #9, opened 2021-09-30
Use dynamic localhost ports in Docker


thekaveman:
See the original writeup in cal-itp/benefits#123 and the implementation in cal-itp/benefits#124.

Relevant docs: https://docs.calitp.org/benefits/getting-started/docker-dynamic-ports/

We should apply the same treatment here.

PR #19, opened 2021-10-04
Refactor app into eligibility_server/app

### LINKED ISSUE #12: Refactor server code into subdirectory
The `app.py` module currently sits at the root of the repository. This is fine for a single file, but can cause trouble when trying to import objects from the module (see e.g. #10) or when the code becomes more complicated than a single file.

Recommendation is to create a subdirectory, named `eligibility_server`, and place both the `app.py` and an (empty) `__init__.py` file in there.

Elsewhere (e.g. in tests) you could then:

```python
from eligibility_server import app
```

machikoyasuda:
## What this PR does
close #12 

- Moves `app.py` into `eligibility_server/app.py`
- Fixes tests by changing `app` import to `eligibility_server.app`
- Uses new method to run Debugger in VS Code: https://code.visualstudio.com/docs/python/tutorial-flask#_run-the-app-in-the-debugger
- Removes `debugpy` dependency

## Testing
Make sure you can:

- [x] - `docker compose build server` `docker compose up server` - open http://localhost:XXXX/healthcheck
- [x] - open in VSCode in remote container: enter command `flask run -h 0.0.0.0` then open http://localhost:XXXX/healthcheck
- [x] - open in VSCode in remote container: get debugger mode running with `F5`
- [x] - open in VSCode in remote container: get debugger mode running with `F5`, change some server-side code, and the debugger should automatically restart 
- [x] - open in VSCode in remove container: run `pytest`

## Docs
image

The new port can be found from the VS Code port panel, or with the docker command: `docker ps`

image


machikoyasuda:
At the moment I'm running this PR's dev container locally with this change:
image

machikoyasuda:
Tried:

```bash
docker-compose run --rm -e DEBUGGER=True -e FLASK_APP=eligibility_server/app.py -e FLASK_ENV=development --service-ports dev python -m debugpy --wait-for-client --listen 0.0.0.0:5678 eligibility_server/app.py
```

and added a breakpoint in 

```python
if __name__ == "__main__":
    print(app)
    print("STUFF")
    breakpoint()
    app.run(host="0.0.0.0", debug=True)  # nosec
```

I was able to get the VS Code debugger working and look into the `app` var, etc., but - once I clicked ▶️, the server broke.


```bash
➜  eligibility-server git:(chore/move-into-folder) ✗ docker-compose run --rm -e DEBUGGER=True -e FLASK_APP=eligibility_server/app.py -e FLASK_ENV=development --service-ports dev python -m debugpy --wait-for-client --listen 0.0.0.0:5678 eligibility_server/app.py
Creating eligibility-server_dev_run ... done
<Flask 'app'>
STUFF
 * Serving Flask app "app" (lazy loading)
 * Environment: development
 * Debug mode: on
 * Running on all addresses.
   WARNING: This is a development server. Do not use it in a production deployment.
 * Running on http://192.168.0.2:5000/ (Press CTRL+C to quit)
 * Restarting with stat
No module named app
```

thekaveman:
So I just came across this: https://code.visualstudio.com/docs/python/tutorial-flask#_run-the-app-in-the-debugger

Which makes this whole thing much easier! I feel like this is new, `debugpy` is built in now!? 🤷 

So two tricks that got this working for me:

**`.vscode/launch.json`** was slightly different from the above docs, but not by much:

```jsonc
{
  "name": "Flask: Eligibility Verification Server",
  "type": "python",
  "request": "launch",
  "module": "flask",
  "args": [
    "run",
    "--host",
    "0.0.0.0"
  ],
  "env": {
    "FLASK_DEBUG": "1",
    "FLASK_ENV": "development"
  }
}
```

This now starts the server on-demand with `F5`, and watch/reload is also active! Possible to start/stop/restart as well.

**`docker-compose`** got much simpler:

```yaml
dev:
  build:
    context: .
    dockerfile: Dockerfile.dev
  command: sleep infinity
  image: eligibility_server:dev
  ports:
    - 5000
  volumes:
    - ./:/home/calitp/app/
```

`command: sleep infinity` ensures the devcontainer continues to run even if the server code crashes/dies, or when debug mode is stopped.

thekaveman:
There's a similar write-up for Django, which I tried with `benefits` and... it worked! 

https://code.visualstudio.com/docs/python/tutorial-django#_create-a-debugger-launch-profile

.vscode/launch.json
```diff
@@ -1,23 +1,24 @@
...
+            "module": "flask",
+            "args": [
+                "run",
+                "--host",
+                "0.0.0.0"
+            ],
+            "env": {
+                "FLASK_DEBUG": "1",
+                "FLASK_ENV": "development",
+                "FLASK_APP": "eligibility_server/app.py"
```
thekaveman, line 20: We can remove the `FLASK_APP` variable since it's already baked into the image.

Dockerfile
```diff
@@ -21,5 +21,7 @@ COPY . /home/$USER/app/
 # switch to non-root $USER
 USER $USER
 
+ENV FLASK_APP=eligibility_server/app.py
```
thekaveman, line 24: Move this variable above with the first `ENV` statement. 1) so they are all grouped together and 2) so the variable is available to later layers in the Docker build process (not as relevant for this variable, but good to keep in mind).

docker-compose.yml
```diff
@@ -12,14 +12,9 @@ services:
     build:
       context: .
       dockerfile: Dockerfile.dev
-    command: python -m debugpy --listen 0.0.0.0:5678 app.py
-    environment:
-      - FLASK_APP=app.py
-      - FLASK_DEBUG=1
-      - FLASK_ENV=development
+    command: sleep infinity
```
thekaveman, line 15: In `benefits` I had to add [`entrypoint: []`](https://github.com/cal-itp/benefits/blob/refactor-debugger/localhost/docker-compose.yml#L37) to ensure the container's entrypoint was null, since it is [set explicitly in the parent Dockerfile](https://github.com/cal-itp/benefits/blob/dev/Dockerfile#L50).

For `eligibility-server` it works without, since the container's entrypoint is already null. But might be a good idea to add it anyway to guard against future breakage if the entrypoint changes.

[Decent answer in SO](https://stackoverflow.com/a/41676550/453168) talking about differences between entrypoint and command.

[Docker docs on `ENTRYPOINT`](https://docs.docker.com/engine/reference/builder/#entrypoint) in Dockerfile; `entrypoint` in docker-compose.yml works the same way. As do `CMD` in Dockerfile and `command` in docker-compse.yml.

docker-compose.yml
```diff
@@ -12,14 +12,9 @@ services:
...
     image: eligibility_server:dev
     ports:
-      - 5000:5000
-      - 5678:5678
+      - 5000
```
thekaveman, line 18: Nice. This partially solves #9!

thekaveman (changes requested):
Great job! Few small changes, added some links to more info about Docker entrypoint/command.

thekaveman (changes requested):
Forgot to add:

You can remove `debugpy` from the [`requirements.txt`](https://github.com/cal-itp/eligibility-server/blob/main/requirements.txt#L2) file too!

thekaveman (approved):
🚀 🎸

PR #22, opened 2021-10-05
WIP - GitHub action that builds and publishes to GitHub Container Registry

machikoyasuda:
### What this PR does
- Create a new GitHub action that runs on pushes to `main`
- The action builds and publishes the `eligibility-server` Docker image to the GitHub Container Registry
- With tags of `latest` for the latest **release** of `main` and a tag of `test` for the latest push to `main`
- The action also filters for pushes to `main` that touch actual app files (like `eligibility_server/app.py` or `requirements.txt`)

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -1,0 +1,41 @@
+name: Build and publish Docker image
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'data/**'
+      - 'eligibility_server/**'
+      - 'keys/**'
+      - 'static/**'
+      - 'tests/**'
+      - '.github/workflows/docker-publish.yml'
+      - 'Dockerfile'
+      - 'requirements.txt'
+
+jobs:
+  build_push:
+    name: Package Docker image
+    runs-on: ubuntu-latest
+    steps:
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push
+        uses: docker/build-push-action@v2
+        if: ${{ github.event_name == 'release' && github.event.action == 'published' }}
+        with:
+          push: true
+          tags: ghcr.io/${{github.repository}}:${{ github.event.release.tag_name }}
+
+      - name: Build and push
+        uses: docker/build-push-action@v2
+        if: ${{ github.event_name != 'release' }}
+        with:
+          push: false
+          tags: ghcr.io/${{github.repository}}:test


.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,43 @@
...
+  release:
+    types: [ published ]
+  push:
+    branches:
+      - main
+    paths:
+      - 'data/**'
+      - 'eligibility_server/**'
+      - 'keys/**'
+      - 'static/**'
```
thekaveman, line 13: I'm going to create a separate ticket for this, but the `static/` folder and the JS file inside can be removed from here. We can talk but this is not part of the server per se.

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,43 @@
...
+      - 'tests/**'
```
thekaveman, line 14: Do we want to do a publish for changes to the tests?

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,43 @@
...
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push
+        uses: docker/build-push-action@v2
+        if: ${{ github.event_name == 'release' && github.event.action == 'published' }}
+        with:
+          push: true
+          tags: ghcr.io/${{github.repository}}:${{ github.event.release.tag_name }}
```
thekaveman, line 36: I like this, using the tag name from the release itself.

In the PR description you mention after a release that the `latest` tag is also published, but I don't see that. Want to add another step just like this one that pushes over the `latest` tag?

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,43 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push
```
thekaveman, line 31: Step name could be more descriptive, esp. to distinguish from below

e.g. `Build and push release`

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,43 @@
...
+        uses: docker/build-push-action@v2
+        if: ${{ github.event_name == 'release' && github.event.action == 'published' }}
+        with:
+          push: true
+          tags: ghcr.io/${{github.repository}}:${{ github.event.release.tag_name }}
+
+      - name: Build and push
```
thekaveman, line 38: See above. Name could be more descriptive e.g. `Build and push test`

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,43 @@
...
+        uses: docker/build-push-action@v2
+        if: ${{ github.event_name != 'release' }}
+        with:
+          push: false
+          tags: ghcr.io/${{github.repository}}:test
```
thekaveman, line 43: Can we call this `main` instead of `test`? That way it corresponds to the branch name and it's clear it is the same as the branch.

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,43 @@
...
+      - 'eligibility_server/**'
+      - 'keys/**'
+      - 'static/**'
+      - 'tests/**'
+      - '.github/workflows/docker-publish.yml'
+      - 'Dockerfile'
+      - 'requirements.txt'
+
+jobs:
+  build_push:
```
thekaveman, line 20: Maybe `build_push_docker` just so it's clear?

thekaveman (changes requested):
Really great start! I requested some changes that will help us read logs and use the action checks in other contexts later if needed. Made a few other suggestions elsewhere.

.github/workflows/docker-publish.yml
machikoyasuda: I looked at this example of how GitHub releases this linter: https://github.com/github/super-linter/blob/main/.github/workflows/deploy-RELEASE-standard.yml#L125 and how it outputs here: https://github.com/github/super-linter/releases

image

I am thinking the action should:

- After pushes to `main` - the `main` package is built and updated
- After a release of `main` at a specific place with a `$version-tag-name` - the package is built and published with 2 tags: the `$version-tag-name`, as well as `latest`

This way, we could - for example:

1. Merge some new server feature in a PR to `main`
2. Test this new server feature *in* the benefits app, pointing to the `main` package
3. Confirm that this `main` package runs well in the benefits app
4. Then, after this is confirmed, *release* the server at this point as a new "release"

machikoyasuda: Haven't tested this new flow out in the fork yet. Will do this first thing tomorrow!

machikoyasuda:
Rested this Action in a fork: https://github.com/machikoyasuda/eligibility-server/pull/8

The steps I took:

1. Merge the PR https://github.com/machikoyasuda/eligibility-server/pull/8
2. After which, the `main` package was updated with this new code
3. After that, I manually created a `release` called `wednesday release` and tagged with `v0.4`

image

PR #32, opened 2021-10-07
Create Database, Verify classes

### LINKED ISSUE #29: Refactor app.py
`app.py` is one giant file with a lot going on. As this project starts to mature, it makes sense to refactor the existing file into a few different modules that can grow and be maintained a little more independently.

Wait until #25 is closed to make it easier!

machikoyasuda:
closes #29 

* move Database class, Verify class into their own files
* add very basic Verify route spec
* creates Settings file that has a list of configs
* create and use Settings variables for hard-coded strings/booleans/numbers

### DIFF
--- a/eligibility_server/app.py
+++ b/eligibility_server/app.py
@@ -22,5 +22,5 @@ api = Api(app)
 api.add_resource(Verify, "/verify")
 
 
-if __name__ == settings.__name__:
+if __name__ == "__main__":
     app.run(host=settings.HOST, debug=settings.DEBUG_MODE)  # nosec
--- a/eligibility_server/settings.py
+++ b/eligibility_server/settings.py
@@ -2,7 +2,6 @@
 App settings for server.
 """
 
-__name__ = "__main__"
 APP_NAME = "eligibility_server.app"
 DEBUG_MODE = True
 HOST = "0.0.0.0"  # nosec


eligibility_server/verify.py
```diff
@@ -0,0 +1,118 @@
...
+            return str(ex), 400
+
+        if token_payload:
+            try:
+                # craft the response payload using parsed request token
+                sub, name, eligibility = token_payload["sub"], token_payload["name"], list(token_payload["eligibility"])
+                resp_payload = dict(
+                    jti=token_payload["jti"],
+                    # iss=app.name,
+                    iss="THIS IS STUBBED",
```
machikoyasuda, line 102: I need to figure out how to either import `app` into this file from `eligibility_server.app` or just import `app.name`.

Currently getting this error when trying to `import eligibility_server.app`:

```bash
$ coverage run -m pytest -q --disable-warnings
ImportError while loading conftest '/home/calitp/app/tests/conftest.py'.
tests/conftest.py:3: in <module>
    from eligibility_server.app import app
eligibility_server/app.py:7: in <module>
    from eligibility_server.verify import Verify
eligibility_server/verify.py:13: in <module>
    from eligibility_server.app import app_name
E   ImportError: cannot import name 'app_name' from partially initialized module 'eligibility_server.app' (most likely due to a circular import) (/home/calitp/app/eligibility_server/app.py)
```

eligibility_server/app.py
```diff
@@ -1,179 +1,23 @@
...
-    @property
-    def jwe_encryption_alg(self):
-        return self._config["jwe_encryption_alg"]
-
-    @property
-    def jws_signing_alg(self):
-        return self._config["jws_signing_alg"]
-
+from flask_restful import Api
+from eligibility_server.verify import Verify
```
thekaveman, line 7: It's conventional to a put a blank line between each set of imports, in the following order:

```python
# system libraries
import datetime
import json

# vendor libraries
import flask
import jwcrypto

# local project modules
from . import settings
from .verify import Verify
```

Notice the syntax on those final local imports; when importing from within the same directory, they can be a bit of a shortcut.

eligibility_server/app.py
```diff
@@ -1,179 +1,23 @@
...
+from eligibility_server.settings import APP_NAME
 
 app = Flask(__name__)
 api = Api(app)
```
thekaveman, line 11: Let's move the `api` creation down after the healthcheck, closer to its resource registration. This way:

1. The `app` is created and configured first
2. Then the `api` is created and configured second

And each step/phase is grouped together in the code.

Also, it is conventional for module-level statements to be separated by double blank lines to make them stand apart from other module-level stuff. [`black`](https://black.readthedocs.io) and the pre-commit hooks should do most of this automatically, but it misses things sometimes. 

```python
"""
Docstring
"""

import library


module_var1 = something()
module_var1.setting = "1"


module_var2 = something_else()


def func():
    # do something


class Cls:
    # define me


if __name__ == "__main__":
    # etc.
```

eligibility_server/app.py
```diff
@@ -1,179 +1,23 @@
...
-                return str(ex), 500
-        else:
-            return "Invalid token format", 400
-
-
 api.add_resource(Verify, "/verify")
 
-
 if __name__ == "__main__":
     app.run(host="0.0.0.0", debug=True)  # nosec
```
thekaveman, line 23: We can move both of these hardcoded values to `settings`!

eligibility_server/app.py
machikoyasuda: I _think_ this next commit resolved this: https://github.com/cal-itp/eligibility-server/pull/32/commits/feba866630f0608f37015a1877c0c38168278f5f

eligibility_server/app.py
```diff
@@ -4,20 +4,23 @@
...
 
 
+api = Api(app)
 api.add_resource(Verify, "/verify")
 
-if __name__ == "__main__":
-    app.run(host="0.0.0.0", debug=True)  # nosec
+
+if __name__ == settings.__name__:
+    app.run(host=settings.HOST, debug=True)  # nosec
```
thekaveman, line 26: One more! `debug=True` 😄

eligibility_server/app.py
thekaveman: yep! perfect

eligibility_server/app.py
```diff
@@ -1,179 +1,26 @@
...
-            return "Invalid token format", 400
-
-
+api = Api(app)
 api.add_resource(Verify, "/verify")
 
 
-if __name__ == "__main__":
-    app.run(host="0.0.0.0", debug=True)  # nosec
+if __name__ == settings.__name__:
```
thekaveman, line 25: Sorry, I'm being pedantic now... 

`"__main__"` is the value of the module's `__name__` variable given by Python, when the module file is run from the command line a la `python file.py`. It doesn't need to be a setting as it's configured by Python automatically and doesn't change.

thekaveman (changes requested):

PR #33, opened 2021-10-13
Move server config to settings

### LINKED ISSUE #30: Refactor Database
The current `Database` class is a simple wrapper around the `server.json` file.

The public API `check_user(key, user, types)` method is great from a consumer point of view. But the internals are tied too closely to the `server.json` file structure.

The database may proxy to a number of data sources:

* `server.json` like file with nested data
* `.csv` file of users in a flat table
* either of the above, but from an S3 or GCP bucket
* real database connection
* etc.

We should refactor to allow different use-cases, perhaps through inheritance or some plug-in mechanism.

machikoyasuda:
closes #30 

closes half of the ticket

### DIFF
--- a/tests/test_settings.py
+++ b/tests/test_settings.py
@@ -18,6 +18,4 @@ def test_settings():
 
 def test_debug():
     if settings.DEBUG_MODE:
-        assert settings.DEBUG_MODE
-    else:
-        assert False
+        assert True


eligibility_server/verify.py
```diff
@@ -11,6 +11,12 @@
 from jwcrypto import jwe, jwk, jws, jwt
 from eligibility_server.database import Database
 from eligibility_server.settings import APP_NAME
+from eligibility_server.settings import AUTH_HEADER
```
thekaveman, line 14: For the import/usage of these settings, I think the pattern we want to go for is more along the lines of:

```python
from . import settings

# etc.

something = settings.APP_NAME
```

This way gets Intellisense on typing `settings.` to see what settings are available, and we'll greatly cut down on the number of "things" we have to import.

thekaveman (approved):
May want to add specs for the `settings` module at some point too.

Issue #34, opened 2021-10-13
Implement hashed lookup


thekaveman:
Building on #30 and [`cal-itp/hashfields`](/cal-itp/hashfields), we want to support (at least) two distinct modes of verification lookup from our source:

1. Normal lookup: does `(id, name)` exist on the list?
2. Hashed lookup: does `(hash(id), hash(name))` exist on the list?

_(note in either case, the Eligibility Request coming into the server will contain the un-hashed data)_

We'll want to be able to configure:

* What kind of lookup (normal or hashed)
* The hash function to use (`sha256`, `sha384`, `sha512`, others?)

PR #51, opened 2021-12-15
Refactor check_user to optionally hash key and user inputs

machikoyasuda:
#34 

## What this PR does
- Refactor `Hash`, `Database` and `Verify` classes
- Add `.env`, `.env.sample`, `.env.tests` file with new env variable, `HASH_TYPE` and default sample value, `sha256`.
- Add a check to  the `Database#check_user` that checks to see if the `Database` has a `Hash` or not. If the `Database` was initialized with a `Hash` object, the `check_user` function will hash the inputs first. If not, the function will not hash the inputs.
- Add hashed data (sha256 and sha512) to `server.json` as test data
- Add test cases
- Add docstring documentation and typing for methods

## To Do

- [x] Refactor hash variables into its own config
- [x] Hash class should read hash vars from config
- [x] Refactor hash vars out of Database class method

## How to run tests

`coverage run -m pytest -v`


eligibility_server/database.py
```diff
@@ -11,14 +12,45 @@ def __init__(self):
...
+        Return a hash of inputted string
+
+        @param input: string
+        @param hash_type: optional, string of hashlib available algorithms defaults to 'sha256'
+        options: 'sha256', 'sha384', 'sha512' and others
+
+        @return string - hashed string
+        """
+
+        return hashlib.new(hash_type, input.encode("utf-8")).hexdigest()
```
machikoyasuda, line 26: Move to new class

eligibility_server/database.py
```diff
@@ -11,14 +12,45 @@ def __init__(self):
...
+        @return array - empty array or array of strings of types user is eligible for
+        """
+
+        key_to_check = key
+        user_to_check = user
+
+        if hash_inputs:
+            key_to_check = self.hash_input(key, hash_type)
+            user_to_check = self.hash_input(user, hash_type)
+
```
machikoyasuda, line 47: Move to same new class

eligibility_server/database.py
```diff
@@ -11,14 +12,45 @@ def __init__(self):
...
+        @param input: string
+        @param hash_type: optional, string of hashlib available algorithms defaults to 'sha256'
+        options: 'sha256', 'sha384', 'sha512' and others
+
+        @return string - hashed string
+        """
+
+        return hashlib.new(hash_type, input.encode("utf-8")).hexdigest()
+
+    def check_user(self, key, user, types, hash_inputs=False, hash_type="sha256"):
```
machikoyasuda, line 28: Move hash_inputs and hash_type to env var / settings.py, to be read within the method

machikoyasuda: Use https://docs.python.org/3/library/configparser.html

machikoyasuda: Parse the env/settings into a class that takes care of configuration

eligibility_server/database.py
```diff
@@ -11,14 +12,45 @@ def __init__(self):
             data = json.load(f)
             self._users = data["users"]
 
-    def check_user(self, key, user, types):
-        """Check if the data matches a record in the database."""
+    def hash_input(self, input, hash_type):
```
machikoyasuda, line 15: Try https://docs.python.org/3/library/typing.html

eligibility_server/database.py
machikoyasuda: Try https://docs.python.org/3/library/typing.html

machikoyasuda:
@thekaveman 
I've tried several ways to mock the environment variables (`HASH_INPUTS` (Bool) and `HASH_TYPE` (String)), but they haven't led to what I expect would happen.

I'm running into several issues:
1. Using `monkeypatch` to directly mock HASH_INPUTS and HASH_TYPE  like this, works -- but the last 2 lines will fail. 

```python
    monkeypatch.setenv("HASH_INPUTS", "False")
    assert os.environ.get("HASH_INPUTS") == "False"
    assert settings.HASH_INPUT == False ## this fails
    assert hash._hash_inputs == False   ## this fails
```

It seems like `monkeypatch` really only makes `os.environ.get()` return the mocked value (`False`). This, however, isn't forcing the `settings` class to be re-run and the `hash` instantiation to then get this updated value of `False`. It appears that both the `settings.py` file and `hash.py` class are still using the _actual_ `.env` var value, of `True`. The way I have the `hash#__init__` method structured, the init method is reading from `settings`.

```
    def __init__(self):
        """
        Initialize variables (hash_inputs, hash_type) from settings.py variables
        """

        self._hash_inputs = settings.HASH_INPUTS or False
        self._hash_type = settings.HASH_TYPE or "sha256"
```

_So_, I tried a different way:

2.
```python
@patch("eligibility_server.settings.HASH_INPUTS", False)
@patch("eligibility_server.settings.HASH_TYPE", "sha512")
```

But I wasn't able to get this to pass:

```
assert settings.HASH_TYPE == "sha512" # This fails
```

I wasn't able to use `@patch` to monkeypatch the `hash` (self) directly either.

3. Then I tried a new way all together, I made a new test directory, with a new `conftest.py` file and created a new _fixture_ that monkeypatches:

image

The good news is that, with this way, I was able to get this test, which is testing the case in which a user doesn't want to hash anything, to pass:

image

However, when I tried to write another spec that tests this same condition (HASH_INPUTS is False) from the `check_user` Database class method, this test failed. When I `print` outputted `self._hash_inputs`, it logged as `True`.... which clued me into think that somehow, still, this variable is getting overridden, or not being mocked correctly:

image

Any suggestions???

I need to be able to:

- Write tests to confirm that changing `.env` variables will instantiate a Hash class correctly. If the `.env` var file says that `HASH_INPUTS` is False, `self._hash_inputs` should return `False`.
- Write tests to confirm that the `hash_input()` method correctly will *not* hash inputs if `self._hash_inputs` is False.
- Write tests to where `hash_type` is set to `sha512` or something else.

I'm not sure if monkeypatching the environment variables are better, or if using the fixture method is better.. (though I'd like to make it easier to turn on/off fixtures or use only certain fixtures with certain tests, so I can more easily change "hash_type" for just a few specs).

machikoyasuda:
@thekaveman Can you review the way I've added the .env vars, config default values and initialized the Hash class?

machikoyasuda:
Create a hash instance here https://github.com/cal-itp/eligibility-server/blob/main/eligibility_server/verify.py#L24

machikoyasuda:
- Settings: Test settings with monkey patching (test to see if defaults are working)
- Hash: Test should instantiate a specific Hash
- Database: Test should use a fixture Hash

thekaveman:
I think you captured most of what we talked about in your notes. A few additional things:

* Hash: `__init__()` should accept the hash function (`SHA256` etc.) as a parameter, rather than reading from settings
* Database: `__init__()` should accept a `Hash` object as a parameter so it is not creating its own object
* Verify: create the Hash and Database objects using `settings.py`

And one additional idea as I was typing this up: I think the `Hash` object can be simplified to not care about `HASH_INPUTS`, instead assuming that if it is being used, you always want to hash the inputs. The settings check can move to `Verify`, which will or won't init a Database with a Hash object based on the setting. Here is psuedocode:

```python
class Database:
    def __init__(self, hash=False):
        self._hash = hash

   def lookup(self, input):
        if self._hash:
             # hash input, then lookup the hashed value
        else:
             # look up input directly


class Verify:
    def __init__(self):
        if settings.HASH_INPUTS:
            hash = Hash(settings.HASH_TYPE)
            self._db = Database(hash=hash)
        else:
            self._db = Database()
```

eligibility_server/hash.py
```diff
@@ -0,0 +1,25 @@
...
+import hashlib
+
+
+class Hash:
+    def __init__(self, hash_type="sha256"):
+        """
+        Initialize variable (hash_type) from settings.py variables
+        """
+
+        self._hash_type = hash_type or "sha256"
```
machikoyasuda, line 14: @thekaveman @angela-tran 

Any thoughts on this design here? It feels off to me.

The logic has now been updated to just have 1 variable (HASH_TYPE). If HASH_TYPE is "", it means the user does not want to hash at all. Any strings in HASH_TYPE will mean hashing will happen. (So now, there's no case in which HASH_INPUTS is True but HASH_TYPE is somehow blank).

Should I remove `="sha256"` and `or "sha256"` all together, and instead, add some sort of validation that checks that `hash_type` is one of the valid hash type strings (`{'sha3_256', 'sha256', 'shake_128', 'sha224', 'sha384', 'mdc2', 'sha3_512', 'ripemd160', 'md5', 'sm3', 'sha512', 'blake2s', 'blake2b', 'whirlpool', 'md5-sha1', 'sha3_224', 'sha512_224', 'md4', 'sha1', 'sha512_256', 'sha3_384', 'shake_256'})`)?

thekaveman: > Should I ... add some sort of validation that checks that `hash_type` is one of the valid hash type strings...

Yes! And raise an exception if it is not. The user of the Server should know right away they have misconfigured it.

This is how `hashfields` does it in [`StringHasher`](https://github.com/cal-itp/hashfields/blob/main/src/Data/StringHasher.cs): with a [list of allowed types](https://github.com/cal-itp/hashfields/blob/main/src/Data/StringHasher.cs#L13), and a [validation check in the init](https://github.com/cal-itp/hashfields/blob/main/src/Data/StringHasher.cs#L36).

machikoyasuda: done!

machikoyasuda:
@thekaveman @angela-tran Finally ready for review. Thank you both for all the help already!

eligibility_server/database.py
```diff
@@ -6,19 +6,44 @@
 
 
 class Database:
-    def __init__(self):
+    def __init__(self, hash=False):
+        """
+        Initialize database with server data and set hash_inputs as True or False
```
thekaveman, line 11: I think this comment should be updated to incorporate the one from below, something like:

> Initialize database with server data and optionally lookup by hashing inputs.

Try not to leak the implementation into the comment ("set hash_inputs as True or False"), becomes harder to keep up to date.

eligibility_server/database.py
```diff
@@ -6,19 +6,44 @@
...
+
+        @param hash: Hash object or False
```
thekaveman, line 13: Here's where this param can be described - why would someone want to pass a Hash object or False?

> @param hash: Hash object to lookup hashed inputs. False to lookup raw inputs.

Something like that.

eligibility_server/database.py
```diff
@@ -6,19 +6,44 @@
...
+
         with open("data/server.json") as f:
             data = json.load(f)
             self._users = data["users"]
 
-    def check_user(self, key, user, types):
-        """Check if the data matches a record in the database."""
+    def check_user(self, key: str, user: str, types: str) -> list:
+        """
+        Check if the data matches a record in the database, optionally, check with hashed key and user strings.
```
thekaveman, line 24: Here I would remove the "optionally, check with hashed key and user strings" - since a user of this Hash object has already configured it. Maybe this is from before the refactor?

eligibility_server/database.py
```diff
@@ -6,19 +6,44 @@
...
+        @return list of strings of types user is eligible for, or empty list
+        """
+
+        if self._hash:
+            key_to_check = self._hash.hash_input(key)
+            user_to_check = self._hash.hash_input(user)
+        else:
+            key_to_check = key
+            user_to_check = user
+
```
thekaveman, line 40: Total nitpick but it might be simpler to write:

```python
if self._hash:
    key = self._hash.hash_input(key)
    user = self._hash.hash_input(user)
```

Or even (to keep some additional verbosity if desired):

```python
key_to_check = key
user_to_check = user

if self._hash:
    key_to_check = self._hash.hash_input(key)
    user_to_check = self._hash.hash_input(user)
```

eligibility_server/hash.py
```diff
@@ -0,0 +1,26 @@
...
+Hashing tool.
+"""
+
+import hashlib
+
+
+class Hash:
+    def __init__(self, hash_type: str):
+        """
+        Initialize variable (hash_type) from settings.py variables
```
thekaveman, line 11: The Hash object shouldn't know where this variable came from (hence its comment shouldn't). Try not to leak implementation in the comment.

eligibility_server/settings.py
```diff
@@ -14,3 +16,7 @@
 JWE_CEK_ENC = "A256CBC-HS512"
 JWE_ENCRYPTION_ALG = "RSA-OAEP"
 JWS_SIGNING_ALG = "RS256"
+
+# Hash Configs from .env file
+
+HASH_TYPE = os.environ.get("HASH_TYPE", "")
```
thekaveman, line 22: Another nitpick, but with broader implications because this name `HASH_TYPE` shows up everywhere (environment variables, settings, compose file, etc.).

It may be more clear, and would line up with the other settings, to call it a `HASH_ALG(ORITHM)` or `HASH_FUNC(TION)`. 

And even more clear would be something like `INPUT_HASH_ALG(ORITHM)/FUNC(TION)` to be explicit it is for the inputs.

tests/test_database.py
```diff
@@ -2,32 +2,114 @@
...
-    DATA = json.load(file)
 
+def test_database_init():
+    hash = Hash("sha256")
+    database = Database(hash)
 
-def test_database_init(database):
     assert database._users
+    if database._hash:
+        assert True
```
thekaveman, line 15: This test feels strange for a few reasons:

1. Typically we want to avoid testing the internals (`database._hash`, `database._users`), especially in duck-typed languages like Python where the public interface / "can it do the thing?" is the most important.
2. The `if` block for `database._hash` is odd; I would expect the test verify that the value passed was set correctly or something similar
3. `assert True` will always pass, it doesn't add a lot of value to the test

tests/test_database.py
```diff
@@ -2,32 +2,114 @@
...
 
-def test_database_check_user_in_database(database):
-    key = min(DATA["users"])
-    user = DATA["users"][key][0]
-    types = DATA["users"][key][1]
+def test_database_init_default():
+    database = Database()
+
+    assert database._users
+    assert database._hash == False  # noqa: E712
```
thekaveman, line 22: Similar to above - this is testing the internals. It's not _necessarily_ a bad thing.

tests/test_database.py
```diff
@@ -2,32 +2,114 @@
...
+
+def test_database_check_user_in_database_with_hashing_specific_type():
+    key = "D4567891"
+    user = "James"
+    types = ["type1"]
+    database = Database(Hash("sha512"))
+
+    response = database.check_user(key, user, types)
+
+    assert response == types
```
thekaveman, line 115: This is a good test. I think we need one more that tests looking up using the wrong hash function configured. I.e. the value was originally hashed with `sha512` but the lookup is configured for `sha256` should return `[]`.

tests/test_database.py
```diff
@@ -2,32 +2,114 @@
...
+
+
+def test_database_check_user_not_in_database():
+    database = Database()
+
+    response = database.check_user("G7778889", "Thomas", ["type1"])
+
+    assert response == []
+
+
```
thekaveman, line 65: All the tests below that use `Hash` verify that `Database` correctly does hashed lookup, and these are great tests! This behavior is ultimately more important than the internal variables that get set in a class/object.

tests/test_database.py
```diff
@@ -2,32 +2,114 @@
 Test database class and methods
 """
 
-import json
+from eligibility_server.database import Database as Database
+from eligibility_server.hash import Hash as Hash
```
thekaveman, line 6: I'm curious about the `import X as X` - that seems redundant vs. `import X`?

tests/test_hash.py
```diff
@@ -0,0 +1,50 @@
+"""
+Test hash class method
+"""
+import pytest
+
+
+from eligibility_server.hash import Hash as Hash
```
thekaveman, line 7: Same as above, `import X as X` ?

thekaveman (changes requested):
This is looking great! A few testing comments, a few comments on the comments 😉, a few more general/python comments.

tests/test_database.py
machikoyasuda: use `is` instead https://www.flake8rules.com/rules/E712.html

tests/test_database.py
machikoyasuda: added

machikoyasuda:
@thekaveman Ready for a re-review:

- renamed `hash_type` to `input_hash_algo`; `HASH_TYPE` to `INPUT_HASH_ALGO`
- added test case, when the wrong hash algorithm is specified, the database should return `[]`
- simplified `check_user` method
- reworded some comments

eligibility_server/verify.py
```diff
@@ -21,8 +22,13 @@
 
 
 class Verify(Resource):
-
-    db = Database()
+    def __init__(self):
+        """Initialize Database with hashing or no hashing"""
```
thekaveman, line 26: Nitpick on this comment - this is the docstring for the `Verify` class init, but the comment speaks to initializing the `Database` class.

tests/test_settings.py
```diff
@@ -16,6 +16,10 @@ def test_settings():
     assert settings.JWS_SIGNING_ALG == "RS256"
 
 
+def test_hash_settings():
+    assert settings.INPUT_HASH_ALGO == "sha256"
```
thekaveman, line 20: We can probably leave this for another ticket - but these tests are where you might want to do some of the environment monkeypatching to ensure settings variables get the correct values from the environment.

eligibility_server/verify.py
machikoyasuda: How's this? https://github.com/cal-itp/eligibility-server/pull/51/commits/33a5d7791a9f32eb0ab42214f86df5dc0efb06be#diff-2537d4df62faed6ef3b3a9fba58d061d10d3fff76e3566faa2fe82b541e701deR26

thekaveman (approved):
Not a blocker in the least, but perhaps a quick `git rebase -i main` could allow for clean up of some of these "trying this" type commits.

PR #53, opened 2022-01-12
chore: configure vscode test debugging

machikoyasuda:
Enabling these settings allows the developer to use VS Code's native testing tools to run the test suite in this application.

image

More on VSCode Python testing configuration: https://code.visualstudio.com/docs/python/testing#_test-configuration-settings


thekaveman (approved):
Wow, this is awesome! Works great in my VS Code:

image

PR #55, opened 2022-01-14
Add settings test

### LINKED ISSUE #54: Settings tests should test different environments / values
These tests are where you might want to do some of the environment monkeypatching to ensure settings variables get the correct values from the environment etc.

_Originally posted by @thekaveman in https://github.com/cal-itp/eligibility-server/pull/51#discussion_r784249237_

machikoyasuda:
closes #54 

## What this PR does:
- Adds 1 test to the `test_settings` file - this test should test that a new `.env` file with a different `INPUT_HASH_ALGO` variable will appropriately set that new value to `settings.INPUT_HASH_ALGO`.
- Adds 1 new `.env.settingstest` file to be used on the new test, with the custom marker, `@pytest.mark.settingstest`
- Adds 1 new `pytest.ini` file to register the new custom marker
- Changes the `run_test` GitHub Action to run the test suite twice: on the first run, test all but the `@pytest.mark.settingstest` spec, and on the second run, run only that spec. In between the two runs, run `source tests/.env.settingstest` so that the test environment can read from the 2nd .env file
- Refactors `test_database` and `test_hash` with parameterization

## How I'm using the custom `@pytest.mark`
1. Register the mark in `pytest.ini` - to document the mark and what it means. This isn't required for implementation, but is highly recommended:

```ini
[pytest]
markers =
    settingstest: mark a test to be run with .env.settingstest
```

2. Use the custom mark in a method: `@pytest.mark.settingstest`:

```python
@pytest.mark.settingstest
def test_hash_settings_env():
    assert settings.INPUT_HASH_ALGO == "sha512"
```

3. Call `pytest` using the mark:

```bash
## Run all test with the mark
coverage run -m pytest -m settingstest
```

```bash
## Run all tests without the mark
coverage run -m pytest -m "not settingstest"
```

This is what the output looks like:

```bash
## Run all test with the mark
======================= 1 passed, 23 deselected in 0.10s =======================
```

```bash
## Run all tests without the mark
======================= 23 passed, 1 deselected in 2.09s =======================
```

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -27,7 +27,6 @@ jobs:
           source tests/.env.settingstest
           coverage run -m pytest -m settingstest
           coverage report -m
-          set +a
           set -a
           source tests/.env.tests
           coverage run -m pytest -m "not settingstest"
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -24,11 +24,11 @@ jobs:
       - name: Test with pytest
         run: |
           set -a
-          source tests/.env.tests
-          coverage run -m pytest -m "not settingstest"
-          set +a
-          set -a
           source tests/.env.settingstest
           coverage run -m pytest -m settingstest
           coverage report -m
+          set +a
+          set -a
+          source tests/.env.tests
+          coverage run -m pytest -m "not settingstest"
           set +a
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -3,12 +3,6 @@ name: Run pytests
 on:
   pull_request:
     branches: [ "*" ]
-    paths:
-      - 'data/**'
-      - 'eligibility_server/**'
-      - 'keys/**'
-      - 'tests/**'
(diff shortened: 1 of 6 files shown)


.github/workflows/run-tests.yml
```diff
@@ -24,7 +24,11 @@ jobs:
       - name: Test with pytest
         run: |
           set -a
-          source tests/.env.tests
-          coverage run -m pytest
+          source tests/.env.settingstest
+          coverage run -m pytest -m settingstest
           coverage report -m
           set +a
+          set -a
```
thekaveman, line 31: It seems redundant to `set +a` then `set -a` right away... maybe just leave the final `set +a` for the end?

thekaveman (changes requested):
I'm liking the parameterized tests!

angela-tran (commented):
The parameterization and the custom marks look good to me!

I agree with @thekaveman 's feedback that we could remove those redundant lines in the workflow.

thekaveman (approved):
🚀

angela-tran (approved):
👍

PR #58, opened 2022-01-25
feat(db): add sql-alchemy, user model, populate database table

machikoyasuda:
WIP

### DIFF
--- a/eligibility_server/app.py
+++ b/eligibility_server/app.py
@@ -2,8 +2,11 @@
 Simple Test Eligibility Verification API Server.
 """
 
+import json
+
 from flask import Flask
 from flask_restful import Api
+from flask_sqlalchemy import SQLAlchemy
 
 from . import settings
 from .verify import Verify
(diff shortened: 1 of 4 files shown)


.github/workflows/run-tests.yml
```diff
@@ -19,8 +19,12 @@ jobs:
         run: |
           set -a
           source tests/.env.settingstest
+          python setup.py
```
angela-tran, line 22: This line is actually necessary before running `databasetest`, correct? I'm just wondering why you have it called before running `settingstest` when it seems that the database isn't needed for `settingstest`.

machikoyasuda: I tried this _but_, because the Database test data requires `Database()` to be initialized, and the `init` method for the Database class now requires the db -- I still have to initialize the database first before running any tests. So for the time being, I am going to keep this `setup.py` up here.

machikoyasuda: This line: https://github.com/cal-itp/eligibility-server/pull/58/files#diff-82fbc96aa3f082d4667c09a2a35aec050c73120de9174e1f37bef26ef9cd3115R12 is where it breaks

angela-tran (approved):
Very nice! :tada:

setup.py
```diff
@@ -0,0 +1,35 @@
+import csv
+import json
+
+from eligibility_server import app, settings
+
+db = app.db
+User = app.User
+
+
+def import_users(db, User):
```
thekaveman, line 10: I'm confused, what is the difference between the variable `User` on line 7, and this parameter `User` on line 10? Seems like if we have a global variable there is no need to parameterize the functions here?

setup.py
```diff
@@ -0,0 +1,35 @@
...
+            for user in data:
+                save_users(db, User, user, data[user][0], str(data[user][1]))
+    if settings.IMPORT_FILE_FORMAT == "csv":
+        with open(settings.IMPORT_FILE_PATH, newline="", encoding="utf-8") as file:
+            data = csv.reader(file, delimiter=";", quotechar="", quoting=csv.QUOTE_NONE)
+            for user in data:
+                save_users(db, User, user[0], user[1], user[2])
+
+
+def save_users(db, User, user_id, key, types):
```
thekaveman, line 24: Same as above - what is `User` here in the parameter list?

setup.py
```diff
@@ -0,0 +1,35 @@
...
+    item = User(user_id=user_id, key=key, types=types)
+    db.session.add(item)
+    db.session.commit()
+
+
+if __name__ == "__main__":
+    print("Creating table...")
+    db.create_all()
+    print("Table created.")
+    import_users(db, User)
```
thekaveman, line 34: Here `User` is being used as the global variable...

setup.py
```diff
@@ -0,0 +1,35 @@
+import csv
+import json
+
+from eligibility_server import app, settings
+
+db = app.db
+User = app.User
```
thekaveman, line 7: See comments below; I'll also note that elsewhere (e.g. `database.py` ...) the usage is like `app.User` - why the need for a special global variable? I think we should be consistent everywhere in the usage.

teardown.py
```diff
@@ -0,0 +1,11 @@
+from eligibility_server import app
+
+db = app.db
+User = app.User
```
thekaveman, line 4: Same as before - pick a format (`User = app.User` or `app.User`) and be consistent throughout.

thekaveman (changes requested):

PR #94, opened 2022-05-25
Move Docker files into .devcontainer

### LINKED ISSUE #91: Refactor repo with .devcontainer
- Move `Dockerfile`, `devcontainer.json`, `docker-compose.yml`, `bin/pre-commit`, `.env.sample` into a new top-level directory, `/.devcontainer`
- Update configuration file points as necessary
- Update docs as necessary

Why? To align this repository's folder architecture with benefits, eligibility-api and current best practices.

machikoyasuda:
closes #91 

## What this PR does
- Moves Docker and related files into .devcontainer, and updates files accordingly
- Updates docs with instruction to cd into .devcontainer

## How to test
- Run `cd .devcontainer` and then `docker compose --no-cache build server` and then `docker compose --no-cache build dev`
- Reopen the repository in VS Code Remote Container. Ensure that pre-commit is installed and running.
- Run the app in Run and Debug mode
- Run `python setup.py` - Should run successfully
- Run `coverage run -m pytest -m databasetest` - All tests should pass
- Run `source tests/.env.settingstest` and then `coverage run -m pytest -m settingstest` - All tests should pass
- Test that editing files locally works


.devcontainer/compose.yml
```diff
@@ -2,30 +2,30 @@ version: "3.8"
 
 services:
   server:
-    build: .
+    build: ..
```
machikoyasuda, line 5: @thekaveman Are these changes necessary (to add the extra `.`) here and in `volumes:` and `context:`? I was not sure.

thekaveman: Yeah I think you do want these:

* `build: ..` tells Docker to build from the parent folder (e.g. not `.devcontainer/` but the root of the repo). This is what we want.
* `volumes:` with the mapping into `/home/calitp/app` is necessary to map your host machine repo into the devcontainer, so you can edit files and have that be reflected back to your host machine (where the files really live). Maybe not as necessary for `server`.

.devcontainer/compose.yml
```diff
@@ -2,30 +2,30 @@ version: "3.8"
...
       - "5000"
     volumes:
-      - ./:/home/calitp/app/
+      - ../:/home/calitp/app/
   dev:
     build:
-      context: .
-      dockerfile: Dockerfile.dev
+      context: ..
+      dockerfile: .devcontainer/Dockerfile
```
thekaveman, line 15: Yeah `context: ..` here does the same as `build: ..` above. 

But since we're telling Docker to use a different Dockerfile, we have to use `context`.

`build` combines the two concepts and assumes you want to use the `Dockerfile` sitting in the `context` directory.

thekaveman (changes requested):

PR #99, opened 2022-07-25
Feat: configurable sub format

### LINKED ISSUE #57: Make sub format check configurable
https://github.com/cal-itp/eligibility-server/blob/main/eligibility_server/verify.py#L114

This line should be configurable

```python
if re.match(r"^[A-Z]\d{7}$", sub):
```

angela-tran:
Closes #57

### DIFF
--- a/eligibility_server/verify.py
+++ b/eligibility_server/verify.py
@@ -91,7 +91,7 @@ class Verify(Resource):
                 iat=int(datetime.datetime.utcnow().replace(tzinfo=datetime.timezone.utc).timestamp()),
             )
             # sub format check
-            if re.match(r"^[A-Z]\d{7}$", sub):
+            if re.match(settings.SUB_FORMAT_REGEX, sub):
                 # eligibility check against db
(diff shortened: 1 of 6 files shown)


thekaveman:
👀

eligibility_server/verify.py
```diff
@@ -102,25 +124,6 @@ def get(self):
...
-                    resp_payload["eligibility"] = self._db.check_user(sub, name, eligibility)
-                    code = 200
-                else:
-                    resp_payload["error"] = {"sub": "invalid"}
-                    code = 400
-                # make a response token with appropriate response code
-                return self._make_token(resp_payload), code
-            except Exception as ex:
-                return str(ex), 500
+            return self._get_response(token_payload)
```
thekaveman, line 127: I imagine some of this will be cleaned up when `eligibility-api` is introduced. 

It could be useful to split the logic of creating the response token from validating the eligibility / doing that check.

thekaveman (approved):

PR #104, opened 2022-08-03
Fix: idempotent setup and teardown

### LINKED ISSUE #103: Make database setup and teardown idempotent
Currently, an attempt to restart a [`server`](https://github.com/cal-itp/eligibility-server/blob/c5ccde1ff1ddd95f66add1a9076475a4990f48f9/.devcontainer/compose.yml#L4) Docker container will result in an `IntegrityError`:
```
+ python setup.py
Creating table...
Table created.
Importing users from /.devcontainer/server/data.csv
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 1819, in _execute_context
    self.dialect.do_execute(
  File "/usr/local/lib/python3.9/site-packages/sqlalchemy/engine/default.py", line 732, in do_execute
    cursor.execute(statement, parameters)
sqlite3.IntegrityError: UNIQUE constraint failed: user.key
```

image

A separate but similar issue is that an attempt to teardown the database if it has already been torn down results in an `OperationalError`:
```
+ python teardown.py
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 1819, in _execute_context
    self.dialect.do_execute(
  File "/usr/local/lib/python3.9/site-packages/sqlalchemy/engine/default.py", line 732, in do_execute
    cursor.execute(statement, parameters)
sqlite3.OperationalError: no such table: user
```

We should make the `setup` and `teardown` scripts idempotent to prevent these errors.

### Additional context
#101 made it so the app container initializes the database on startup.

angela-tran:
Closes #103

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,7 @@
 import csv
 import json
 
+from flask_sqlalchemy import inspect
 from eligibility_server import app, settings
 
 
@@ -33,8 +34,13 @@ def save_users(user_id: str, key: str, types: str):
 
 
 if __name__ == "__main__":
-    print("Creating table...")
-    app.db.create_all()
-    print("Table created.")
-    import_users()
-    print(app.User.query.count(), "users added.")
+    inspector = inspect(app.db.engine)
+
+    if inspector.get_table_names():
+        print("Tables already exist.")
+    else:
+        print("Creating table...")
+        app.db.create_all()
+        print("Table created.")
+        import_users()
+        print(app.User.query.count(), "users added.")
--- a/setup.py
+++ b/setup.py
@@ -12,11 +12,13 @@ def import_users():
             data = json.load(file)["users"]
             for user in data:
                 save_users(user, data[user][0], str(data[user][1]))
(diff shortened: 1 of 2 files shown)


thekaveman (approved):
Nice find and nice simple fix 👍 

In `setup.py`, should there be one more check to see if, when the table already exists, it still needs users to be loaded? Not sure if that's a use case that matters...

angela-tran:
> In `setup.py`, should there be one more check to see if, when the table already exists, it still needs users to be loaded? Not sure if that's a use case that matters...

I could see a case where maybe the `IMPORT_FILE_PATH` was wrong and the table was created but no data actually went into it. I'll add this check 👍 

I think I'll also add a print statement for the case where data wasn't loaded.

thekaveman (approved):
Looks good!

Maybe a follow-up: figure out how to get the Python `logging` framework configured and replace the `print` statements with more appropriate log-level statements.

PR #105, opened 2022-08-03
CSV configuration variables

### LINKED ISSUE #62: Make CSV options configurable
Currently [when setting up from a CSV input file](https://github.com/cal-itp/eligibility-server/blob/main/setup.py#L14), some settings are assumed, but we likely want them to be configurable:

```python
if settings.IMPORT_FILE_FORMAT == "csv":
  with open(settings.IMPORT_FILE_PATH, newline="", encoding="utf-8") as file:
    data = csv.reader(file, delimiter=";", quotechar="", quoting=csv.QUOTE_NONE)
```

* `delimiter` may not be a semicolon `;`
* values may be quoted
* we may want to handle newlines differently depending on the file origin

Let's make these settings that can be configured from the environment.

machikoyasuda:
closes #62 

## What this PR does

Adds the following environment variables to allow developers to configure the CSV file. Updates .env.sample and .env.settingstest to reflect new variables. Configures values to be compatible with `python setup.py` and the pytests. Updates documentation.

Up next: Refactoring the way settings are configured, #109, so that it is easier to write pytests for different configuration scenarios.

### Environment variables
- CSV_NEWLINE=""
- CSV_DELIMITER=;
- CSV_QUOTING=3
- CSV_QUOTECHAR=""

These variables come straight from https://docs.python.org/3/library/csv.html#examples


angela-tran:
@machikoyasuda and I learned today that `;` is a control operator in bash, so we need to wrap it in quotes when setting it as an environment variable.

https://stackoverflow.com/a/67507121

thekaveman (approved):
This looks great, nice work @machikoyasuda 👍

docs/getting-started.md
```diff
@@ -63,39 +79,39 @@ This repository comes with a [VS Code Remote Containers](https://code.visualstud
 Once you clone the repository locally, open it within VS Code, which will prompt you to re-open the repository within the Remote Container.
 
  1. Build and Open the Dev Container
- 2. Start the `eligibility-server` app with `F5`
- 3. Run any database scripts and test commands from within the container
+ 2. Start the `eligibility-server` Flask app and database with `F5`
+ 3. Now you can run tests from the container.
 
-## Set up database and run tests
+Starting the Dev Container will run `bin/start.sh`, which runs `setup.py` and starts the Flask app. The `setup.py` script creates a table, imports and saves users from a JSON or CSV file specified in the .env file from the `IMPORT_FILE_PATH` key. CSV files will require
```
angela-tran, line 85: Thank you for updating these docs! Especially with the part about `bin/start.sh`, which I missed in #101.

I don't want to block this PR, but I noticed this fragment `CSV files will require`. Maybe we can finish it out or remove it as a follow-up.

angela-tran (approved):

docs/getting-started.md
machikoyasuda: Will take care of this in the next PR.

PR #116, opened 2022-08-08
Refactor: Add logging

### LINKED ISSUE #108: Add logging to server code and setup/teardown scripts
We should configure logging for the project as a whole and add relevant statements throughout the code.

_Originally posted by @thekaveman in https://github.com/cal-itp/eligibility-server/pull/104#pullrequestreview-1061063341_

machikoyasuda:
closes #108 

### What this PR does:
- Configures standard Flask logging in the app
- Adds a log for healthcheck
- Uses Python logging for the setup/teardown scripts

## What it looks like:

Go to /healthcheck route
Check logs
```
[2022-08-08 21:34:43,106] INFO in app: Healthcheck
```

Run the scripts
image

### DIFF
--- a/eligibility_server/app.py
+++ b/eligibility_server/app.py
@@ -5,10 +5,29 @@ Simple Test Eligibility Verification API Server.
 from flask import Flask
 from flask_restful import Api
 from flask_sqlalchemy import SQLAlchemy
+from logging.config import dictConfig
 
 from . import settings
 from .verify import Verify
 
+dictConfig(
+    {
+        "version": 1,
+        "formatters": {
+            "default": {
+                "format": "[%(asctime)s] %(levelname)s in %(module)s: %(message)s",
+            }
+        },
+        "handlers": {
+            "wsgi": {
+                "class": "logging.StreamHandler",
+                "stream": "ext://flask.logging.wsgi_errors_stream",
+                "formatter": "default",
+            }
+        },
+        "root": {"level": "INFO", "handlers": ["wsgi"]},
+    }
+)
 
 app = Flask(__name__)
 app.name = settings.APP_NAME
@@ -18,6 +37,7 @@ app.config["SQLALCHEMY_TRACK_MODIFICATIONS"] = settings.SQLALCHEMY_TRACK_MODIFIC
 
 @app.route("/healthcheck")
 def healthcheck():
+    app.logger.info("Healthcheck")
     return "Healthy"
 
 
--- a/setup.py
+++ b/setup.py
@@ -3,6 +3,7 @@ import json
 
 from flask_sqlalchemy import inspect
 from eligibility_server import app, settings
+import logging
 
 
 def import_users():
@@ -14,7 +15,7 @@ def import_users():
     configurations: CSV_DELIMITER, CSV_NEWLINE, CSV_QUOTING, CSV_QUOTECHAR
     """
 
-    print("Importing users from", settings.IMPORT_FILE_PATH)
+    logging.info(f"Importing users from {settings.IMPORT_FILE_PATH}")
     if settings.IMPORT_FILE_FORMAT == "json":
         with open(settings.IMPORT_FILE_PATH) as file:
             data = json.load(file)["users"]
@@ -31,9 +32,9 @@ def import_users():
             for user in data:
                 save_users(user[0], user[1], user[2])
     else:
-        print(f"File format {settings.IMPORT_FILE_FORMAT} is not supported.")
+        logging.warning(f"File format {settings.IMPORT_FILE_FORMAT} is not supported.")
 
-    print(app.User.query.count(), "users added.")
+    logging.info(f"{app.User.query.count()} users added.")
 
 
 def save_users(user_id: str, key: str, types: str):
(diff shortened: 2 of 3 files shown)


machikoyasuda:
@thekaveman @angela-tran Any suggestions for anywhere else to add logs in the app?

eligibility_server/app.py
```diff
@@ -5,10 +5,29 @@
...
 
 from . import settings
 from .verify import Verify
 
+dictConfig(
+    {
+        "version": 1,
+        "formatters": {
+            "default": {
+                "format": "[%(asctime)s] %(levelname)s in %(module)s: %(message)s",
```
thekaveman, line 18: Can we mirror the [`benefits` format](https://github.com/cal-itp/benefits/blob/dev/benefits/logging.py#L7)? To get the line number in there too:

```
[{asctime}] {levelname} {name}:{lineno} {message}
```

And date formatting `"datefmt": "%d/%b/%Y %H:%M:%S"`

machikoyasuda: How does this look?

image
image

thekaveman: Yeah that's better! I have some other comments related to `name` but will review again and resolve this one.

eligibility_server/app.py
```diff
@@ -5,10 +5,30 @@
...
+            }
+        },
+        "handlers": {
+            "wsgi": {
+                "class": "logging.StreamHandler",
+                "stream": "ext://flask.logging.wsgi_errors_stream",
+                "formatter": "default",
+            }
+        },
+        "root": {"level": "INFO", "handlers": ["wsgi"]},
```
thekaveman, line 29: We might want to configure this default `level` to be something different depending on the environment (local, dev, production, etc.)

What do you think about moving the entire logging configuration into the `settings.py` module and adding an environment variable [similar to `benefits`](https://github.com/cal-itp/benefits/blob/dev/benefits/settings.py#L224)?

setup.py
```diff
@@ -31,9 +32,9 @@ def import_users():
             for user in data:
                 save_users(user[0], user[1], user[2])
     else:
-        print(f"File format {settings.IMPORT_FILE_FORMAT} is not supported.")
+        logging.warning(f"File format {settings.IMPORT_FILE_FORMAT} is not supported.")
 
-    print(app.User.query.count(), "users added.")
+    logging.info(f"{app.User.query.count()} users added.")
```
thekaveman, line 37: `Users added: {count()}`

teardown.py
```diff
@@ -1,14 +1,15 @@
 from flask_sqlalchemy import inspect
 from eligibility_server import app
+import logging
 
 if __name__ == "__main__":
     inspector = inspect(app.db.engine)
 
     if inspector.get_table_names():
-        print(app.User.query.count(), "users to be deleted.")
+        logging.info(f"{app.User.query.count()} users to be deleted.")
```
thekaveman, line 9: `Users to be deleted: {count()}`

setup.py
```diff
@@ -31,9 +32,9 @@ def import_users():
             for user in data:
                 save_users(user[0], user[1], user[2])
     else:
-        print(f"File format {settings.IMPORT_FILE_FORMAT} is not supported.")
+        logging.warning(f"File format {settings.IMPORT_FILE_FORMAT} is not supported.")
```
thekaveman, line 35: `File format is not supported: {settings.IMPORT_FILE_FORMAT}`

setup.py
```diff
@@ -3,6 +3,7 @@
 
 from flask_sqlalchemy import inspect
 from eligibility_server import app, settings
+import logging
```
thekaveman, line 8: Here we should create a `logger` instance using the `logging.getLogger()` function and the `__name__` property of the module:

```python
import logging

logger = logging.getLogger(__name__)
```

Then use the `logger` instance instead of the `module` reference

```diff
-logging.info("...")
+logger.info("...")
```

teardown.py
```diff
@@ -1,14 +1,15 @@
 from flask_sqlalchemy import inspect
 from eligibility_server import app
+import logging
```
thekaveman, line 4: Same overall comment as in `setup.py`: create a `logger` with `logging.getLogger(__name__)` and use it instead to write messages.

thekaveman (changes requested):
Change to the way logging is configured and where, to enable overriding the `level`.

Some changes to how the logger instance is used, to ensure the `name` property is set correctly.

Few changes to the message formats, to make later searching/parsing easier.

thekaveman (commented):
Additional areas to add logging to (I think these can mostly be `DEBUG` or `INFO` level):

* `database.py:__init__()`, print `hash` if set (optionally print number of `users`)
* `database.py:check_user()`, print when hashing input
* `database.py:check_user()`, if possible we should break up the big `if` and print a log message for each different case/mismatch
* `hash.py:__init__()`: print which hash algorithm is configured
* `verify.py`: lots of places to add different levels of messages

thekaveman:
Hey there @machikoyasuda - I'm reconsidering the `database.py` suggestions, @angela-tran and I were talking about some refactors to how that class is working anyway. If you haven't already done this:

> `database.py:check_user()`, if possible we should break up the big if and print a log message for each different case/mismatch

then don't worry about it for now, we'll grab it in a follow-up (I'm writing it up this morning).

eligibility_server/app.py
machikoyasuda: Going to do this in a separate subsequent PR after #119 is merged

machikoyasuda:
Adds logs to Verify class, but not Database yet - Will happen in #122

thekaveman (approved):
This is a good start to our logging config, nice 👍

angela-tran (approved):

thekaveman:
Since @machikoyasuda is out today, and we want to get this in for a number of other PRs, I'm going to merge.

PR #119, opened 2022-08-10
Use Flask configuration framework

### LINKED ISSUE #109: Use Flask configuration framework to load settings
Instead of getting config manually from the `settings` module like:

```python
app = Flask(__name__)
app.name = settings.APP_NAME
app.config["SQLALCHEMY_DATABASE_URI"] = settings.SQLALCHEMY_DATABASE_URI
# etc.
```

Flask has built-in support for loading the configuration from a number of sources. We're using the [Python files format](https://flask.palletsprojects.com/en/2.2.x/config/#configuring-from-python-files), which would look something like:

```python
app = Flask(__name__)
```

We can also load from a number of [supported file types](https://flask.palletsprojects.com/en/2.2.x/config/#configuring-from-data-files):

```python
import json
app.config.from_file("config.json", load=json.load)
```

Or directly from certain [environment variables](https://flask.palletsprojects.com/en/2.2.x/config/#configuring-from-environment-variables):

```python
app.config.from_prefixed_env()
```

Let's use the built-in framework for loading configuration data.

See also: [Flask Configuration Best Practices](https://flask.palletsprojects.com/en/2.2.x/config/#configuration-best-practices)

angela-tran:
Closes #109 

### Changes
 - Replaces direct usage of `settings` with `app.config` instead
 - Loads settings into `app.config` using `from_object`
 - Calls `app.config.from_envvar("ELIGIBILITY_SERVER_SETTINGS")` to allow optional overrides
 - Refactors tests to use a config fixture which can have its values changed
 - Deletes `test_settings.py` since now we're just using Flask's configuration framework
 - Updates docs

### ⚠️ Devs should update their .env files
Move any custom settings to `config/sample.py`, and change your .env file to simply set `ELIGIBILITY_SERVER_SETTINGS=../config/sample.py`.


angela-tran:
I anticipate needing to resolve merge conflicts with #116

machikoyasuda (approved):

thekaveman:
Taking a look. Will wait for the eventual rebase 😅 before submitting a review.

config/sample.py
```diff
@@ -0,0 +1,6 @@
+IMPORT_FILE_PATH = "data/server.csv"
```
thekaveman, line 1: Can we add an entry to `.gitignore` for other files in this directory? That way we can have local testing configs and not have to worry about git looking at them.

machikoyasuda (changes requested):
One change:

Add `"python.defaultInterpreterPath": "/usr/local/bin/python",` to the VS Code `settings.json` file. 

Changing this allows my app to open in debug mode.

config/sample.py
angela-tran: Updated `.gitignore` in [d86d7ea](https://github.com/cal-itp/eligibility-server/pull/119/commits/d86d7ea3f159faf9d297360bd1f23b2845e6c79c)

thekaveman (approved):

thekaveman (approved):

angela-tran:
Sorry @thekaveman , I hit re-request right as you approved 😅

machikoyasuda (approved):

PR #128, opened 2022-08-16
Configurable log level, plus settings documentation

### LINKED ISSUE #107: Add documentation for SUB_FORMAT_REGEX
#99 introduced an environment variable to configure the expected format for the `sub` field from requests.

The docs should be updated to include this setting.

### LINKED ISSUE #127: Add log level configuration as a setting
We might want to configure this default `level` to be something different depending on the environment (local, dev, production, etc.)

What do you think about moving the entire logging configuration into the `settings.py` module and adding an environment variable [similar to `benefits`](https://github.com/cal-itp/benefits/blob/dev/benefits/settings.py#L224)?

_Originally posted by @thekaveman in https://github.com/cal-itp/eligibility-server/pull/116#discussion_r940706723_

angela-tran:
Closes #127 (configurable log level) and closes #107 (SUB_FORMAT_REGEX docs)

Initially, I only extracted the logging configuration and added documentation for the new `LOG_LEVEL` setting. But then I noticed a lot of our settings were missing, so I went ahead and wrote out docs for all the available settings. I formatted it similarly to Benefit's [Environment variables](https://docs.calitp.org/benefits/configuration/environment-variables/) docs.


eligibility_server/app.py
```diff
@@ -5,34 +5,20 @@
...
-)
-
 app = Flask(__name__)
 app.config.from_object("eligibility_server.settings")
 app.config.from_envvar("ELIGIBILITY_SERVER_SETTINGS", silent=True)
 
+logging.configure(app.config["LOG_LEVEL"])
+# remove the default handler since we configured logging after creating the app
+# https://flask.palletsprojects.com/en/2.2.x/logging/#removing-the-default-handler
+app.logger.removeHandler(default_handler)
```
thekaveman, line 20: Why do we need to remove the default handler? 

Corollary, why are we adding a new handler when it is configured basically the same as the default? Seems like what we are changing is the `formatter`, can we just configure that and not have to override/remove the handler?

angela-tran: Interesting idea, I didn't notice that the formatter was the only difference. I'll try out just modifying the configuration that is set by default on the `app`.

thekaveman (approved):
This looks great! Thank you for the docs too

PR #129, opened 2022-08-17
Raise JSON errors, add app error handling methods and test

### LINKED ISSUE #110: Handle application errors
See [Flask - Handling Application Errors](https://flask.palletsprojects.com/en/2.2.x/errorhandling/#handling-application-errors)

Specifically, we should look at [Returning API Errors as JSON](https://flask.palletsprojects.com/en/2.2.x/errorhandling/#returning-api-errors-as-json).

* `403` (forbidden, not authenticated/authorized)
* `404` (endpoint not found)
* `500` (other application error / Exception)

machikoyasuda:
closes #110 

- Adds JSON error handling for 403, 500 in Verify class (https://flask-restful.readthedocs.io/en/latest/quickstart.html?highlight=abort#full-example)
- Adds global 404 error handling for the whole app with logging (https://flask.palletsprojects.com/en/2.2.x/errorhandling/#returning-api-errors-as-json).
- Adds global 401, 403, 500 error handling methods for the whole app with logging. 
- Adds tests

image
image

404 with logging:
image


angela-tran:
Forgot to comment earlier that I am reviewing 👀

angela-tran (commented):
@machikoyasuda I read through these changes as well as some [Flask](https://flask.palletsprojects.com/en/2.2.x/errorhandling/#returning-api-errors-as-json) / [Flask-RESTful](https://flask-restful.readthedocs.io/en/latest/extending.html?highlight=content-type#custom-error-handlers) docs and had some thoughts:

I noticed that our `Verify` class is a `Resource` from Flask-RESTful. I learned today that Flask-RESTful routes are slightly different than Flask routes, specifically in their default Content-Type and error-handling. I'm currently thinking:
 - we don't actually need to call `jsonify` since the default Content-Type of responses for Flask-RESTful routes is already `application/json`
 - when a Flask-RESTful route throws an exception, it doesn't trigger any Flask error-handlers. The call to `abort` in your PR is working because Flask-RESTful calls[ `handle_error`](https://flask-restful.readthedocs.io/en/latest/extending.html?highlight=content-type#custom-error-handlers) to turn the exception into a `Response` object.

Let me know what you think, and we can talk more / pair tomorrow if needed.

thekaveman (commented):
Thanks for that feedback @angela-tran. 

Further, I want to make sure this PR addresses/handles _global application errors_ (in addition to those that might occur specifically inside the Verify API).

We don't want to return Flask's default error response if e.g. someone hits `/not/a/route` - we want a custom, simple JSON API-like response for the 404.

machikoyasuda:
Ready for re-review @angela-tran @thekaveman 

Changes:
- Refactored the Verify class errors (remove `jsonify`)
- Refactored the Verify tests (test actual API endpoint, rather than the method)
- Add 404 error handling to the App and tests

eligibility_server/verify.py
```diff
@@ -107,7 +108,7 @@ def _get_response(self, token_payload):
             return self._make_token(resp_payload), code
         except Exception as ex:
             logger.warning(f"Internal server error: {ex}")
-            return str(ex), 500
+            abort(500, description=f"Internal server error: {ex}")
```
thekaveman, line 111: Couple of comments here:

* As above, we probably should not send the entire `Exception` here, it could contain sensitive values etc. Rather just send something simple like `{ "error": "Server error" }` or similar
* And actually, would it be possible to send a more nuanced error code here? E.g. 400 if the token couldn't be parsed (that's a client problem if the token was malformed)?

eligibility_server/verify.py
```diff
@@ -121,7 +122,7 @@ def get(self):
             headers = self._check_headers()
         except Exception:
             logger.warning("Unauthorized")
-            return "Unauthorized", 403
+            abort(403, description="Unauthorized")
```
thekaveman, line 125: This looks great with the `description` key! Seems like it turns into a `{ "message": "..." }` based on the test?

thekaveman (commented):
This is looking a lot better! Thanks for adding the top-level `@app.errorhandler(404)`

Should we also have top-level handlers for `400`, `403`, and `500` - just in case?

eligibility_server/verify.py
machikoyasuda: This is what it looks like on the API response side:

image

It's just a JSON that returns a string

machikoyasuda: This is what the response looks like from the Pytest perspective:

image

thekaveman: So for `500` I think we should try to be a little more careful and not send the actual error message. If we can determine what type of error it is / what went wrong, and send a more appropriate response (maybe `400` in some cases) or else a pared down error message that gives a little info but not the full exception string.

eligibility_server/app.py
```diff
@@ -40,6 +40,11 @@ def healthcheck():
     return "Healthy"
 
 
+@app.errorhandler(404)
+def page_not_found(error):
+    return jsonify(error=str(error)), 404
```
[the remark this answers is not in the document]
machikoyasuda: Refactored this to return only strings

eligibility_server/verify.py
machikoyasuda: Tried to resolve this here https://github.com/cal-itp/eligibility-server/pull/129/commits/586d6694885302bdb60e6dd7cbbe3aab19aeaaa6

eligibility_server/verify.py
```diff
@@ -106,8 +107,11 @@ def _get_response(self, token_payload):
             # make a response token with appropriate response code
             return self._make_token(resp_payload), code
         except Exception as ex:
-            logger.warning(f"Internal server error: {ex}")
-            return str(ex), 500
+            logger.warning(f"Error: {ex}")
+            if ex.__class__ == TypeError or ValueError:
```
thekaveman, line 111: This is better, but it would be clearer and more "pythonic" to catch these exceptions specifically in the `except` line, e.g.

```python
try:
   # something
except (TypeError, ValueError) as e:
  # do something with these error types
except Exception e:
  # catch-all error handler
```

Check out https://docs.python.org/3/tutorial/errors.html#handling-exceptions for more info

eligibility_server/app.py
```diff
@@ -40,6 +40,12 @@ def healthcheck():
     return "Healthy"
 
 
+@app.errorhandler(404)
+def page_not_found(error):
+    print(error)
```
thekaveman, line 45: Let's log instead of printing

eligibility_server/app.py
```diff
@@ -40,6 +40,12 @@ def healthcheck():
     return "Healthy"
 
 
+@app.errorhandler(404)
```
thekaveman, line 43: Do we need an `@app.errorhandler()` for `400`, `403` or `500`?

machikoyasuda: I talked with @angela-tran about this to confirm my understanding.

The `@app.errorhandler()` is used for routes that are handled by `Flask` (but not `Flask-RESTful`). The `400`, `403`, `500` error handling right now are all in the `Verify` class which uses `Resource` from `Flask-RESTful`, so we're using the `Flask-RESTful` style of error handling with `abort()`, which does not require `@app.errorhandler()`.

angela-tran: @machikoyasuda I think it could still be helpful to add some `@app.errorhandler(400)`, `@app.errorhandler(403)`, `@app.errorhandler(500)` functions in case some other Flask routes throw those status codes. You're right that `Verify` is the only code currently that raises those, but when #136 is merged in, we will have a new Flask route that could throw errors.

angela-tran: I think these error-handling functions would look very similar to what you have here for 404. One difference of course is that they'd return their own error codes.

eligibility_server/verify.py
```diff
@@ -121,7 +125,7 @@ def get(self):
             headers = self._check_headers()
         except Exception:
             logger.warning("Unauthorized")
-            return "Unauthorized", 403
+            abort(403, description="Unauthorized")
```
angela-tran, line 128: I realize that this PR didn't introduce this `"Unauthorized"` string, but in any case, I suggest changing it to `Forbidden` or maybe just removing it altogether to better align with HTTP standards. I noticed [`403 Forbidden`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403) is slightly different from [`401 Unauthorized`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401).

cc @thekaveman in case you have thoughts on this

eligibility_server/verify.py
```diff
@@ -120,19 +124,19 @@ def get(self):
...
+            abort(403, description="Forbidden")
 
         # parse inner payload from request token
         try:
             token = self._get_token(headers)
             token_payload = self._get_token_payload(token)
         except Exception as ex:
-            logger.warning(f"Bad request: {ex}")
+            logger.error(f"Bad request: {ex}")
             return str(ex), 400
```
thekaveman, line 136: I think this is another area where we shouldn't return the actual exception string. Maybe use the same `abort()` syntax?

eligibility_server/verify.py
```diff
@@ -120,19 +124,19 @@ def get(self):
...
 
         if token_payload:
             return self._get_response(token_payload)
         else:
-            logger.warning("Invalid token format")
+            logger.error("Invalid token format")
             return "Invalid token format", 400
```
thekaveman, line 142: Should we use `abort()` here instead?

thekaveman (commented):
I have just 2 more comments, otherwise this is looking great!

thekaveman (approved):
Very nice!

angela-tran (approved):
Looks good to me

Issue #122, opened 2022-08-10
Refactor database class


thekaveman:
As we move into becoming production-ready, there are a few items of cleanup in the [`Database` class](https://github.com/cal-itp/eligibility-server/blob/main/eligibility_server/database.py) that should happen.

cc @angela-tran who discussed this with me earlier today and informed this ticket. 
cc @machikoyasuda we held off on some of this to make #116 more straightforward, documenting here instead.

| Method | Feedback |
| --------- | ----------- |
| `__init__()` | let's not query and store all users. We may have a large list of users, and don't necessarily want to hold them all in-memory |
| `__init__()` | log the given `hash` option as `DEBUG` |
| `check_user()` | break up large `if(...)` statement checking the various states into `if/elif/else` etc. |
| `check_user()` | query for the user here |
| `check_user()` | separate calculation of the return value from the `return` statement |
| `check_user()` | add various logging to the above refactors |

Let's also verify the tests are providing adequate coverage and scoping.

thekaveman:
I suggest we make 2 tables:

## `user`

Existing, but modified:

| Column | Type | Description |
|----------|------|--------------|
| `id`        | `int` | Autogenerated PK |
| `user_id` | `str` | Imported user id |
| `user_name` | `str` | Imported user name |

## `eligibility`

New:

| Column | Type | Description |
|----------|------|--------------|
| `id`        | `int` | Autogenerated PK |
| `user_id` | `int` | Reference to `user.id` |
| `type` | `str` | Name of eligibility type |

This will give us a nice way to separate loading users from the types. And it puts us on the path to having a default, server-wide type that applies to all users when the data doesn't come with types (e.g. we populate the `eligibility` table with a record for each user with the default type if configured).

Read more about model relationships:

* [`Flask-SQLAlchemy`: One-to-many Relationships](https://flask-sqlalchemy.palletsprojects.com/en/2.x/models/#one-to-many-relationships)
* [`SQLAlchemy`: Basic relationship patters > One to many](https://docs.sqlalchemy.org/en/14/orm/basic_relationships.html#one-to-many)

angela-tran:
I felt a little stuck / deadlocked with #130 and #131, and then I realized it would simplify things if I removed the need to hold all users in memory. Once #132 is merged in, I'll have a path forward for #131.

Issue #131, opened 2022-08-19
Make it easier for tests to use a separate app and database


angela-tran:
This is related to #130 (and its original ticket #122).

I am trying to write tests that cover the cases mentioned in #105:

**<details><summary>Cases**</summary>
> ## Testing
> ### Test data
> 
>     Test JSON without hashing
>     Test JSON with hashing
>     Test CSV with default CSV configs, no hashing
>     Test CSV with custom CSV configs, no hashing
>     Test CSV with custom CSV configs, with hashing
> 
> ### Test scenarios
> 
>     Test a JSON without any hashing
>     Test a JSON with sha256
>     Test a JSON with sha512
>     Test a CSV without any hashing, without any special CSV configs / only default CSV configs
>     Test a CSV with sha256
>     Test a CSV with sha512
>     Test a CSV with custom configs: custom delimiter
>     Test a CSV with custom configs: custom quote characters
>     Test a CSV with custom configs: custom delimiter, and hashing with sha256
>     Test a CSV with custom configs: custom quote characters, and hashing with sha256
</details>

I'm having difficulty writing a test for the `setup` module since it directly imports the `app` object from `eligibility_server.app`. My [attempt](https://github.com/cal-itp/eligibility-server/pull/130/commits/bfc46118c726ee112b28971ac3e59caa1360d864) to patch the module's instance and configure a test database on that patch got me very close, but then I'm not able to query for `User`s or `Eligibility`s because those are defined on the `eligibility_server.app.db` instance.

I think we could do something like what the Flask documentation describes as the ["app factory"](https://flask.palletsprojects.com/en/2.2.x/patterns/appfactories/) pattern.
Some examples I found:
 - https://gitlab.com/patkennedy79/flask_user_management_example/-/blob/main/project/__init__.py#L22
 - https://github.com/rochacbruno/flask-project-template/blob/main/project_name/base.py#L5
    - [conftest.py](https://github.com/rochacbruno/flask-project-template/blob/main/tests/conftest.py#L11-L12)

Additional resources:
 - https://www.patricksoftwareblog.com/testing-a-flask-application-using-pytest/

angela-tran:
I haven't started on this yet, but plan to start tomorrow

thekaveman:
The ["app factory"](https://flask.palletsprojects.com/en/2.2.x/patterns/appfactories/) Flask docs start by saying:

> If you are already using packages and blueprints for your application... 

[...]

PR #132, opened 2022-08-19
Refactor: Database class - remove in memory data

angela-tran:
This PR covers the feedback items in the description of #122:

- Query for user in `check_user` which allows us to remove the `_user` variable that was holding all users in-memory
- Log the `hash` being used, and split out conditions in `check_user` to add logging for each
- Separate out calculation of matching types from return statement

### DIFF
--- a/eligibility_server/database.py
+++ b/eligibility_server/database.py
@@ -5,6 +5,9 @@ Simple hard-coded server database.
 import ast
 
 from . import app
+import logging
+
+logger = logging.getLogger(__name__)
 
 
 class Database:
@@ -17,9 +20,9 @@ class Database:
 
         self._hash = hash
         if hash:
-            app.app.logger.debug(f"Database initialized with hash: {hash}")
+            logger.debug(f"Database initialized with hash: {hash}")
         else:
-            app.app.logger.debug("Database initialized without hashing")
+            logger.debug("Database initialized without hashing")
 
     def check_user(self, key: str, user: str, types: str) -> list:
         """
@@ -44,15 +47,13 @@ class Database:
             existing_user_types = None
 
         if len(types) < 1:
-            app.app.logger.debug("List of types to check was empty.")
+            logger.debug("List of types to check was empty.")
             return []
         elif existing_user is None:
-            app.app.logger.debug(f"Database does not contain requested user with sub, name: {key, user}")
+            logger.debug(f"Database does not contain requested user with sub, name: {key, user}")
             return []
         elif len(set(existing_user_types) & set(types)) < 1:
-            app.app.logger.debug(
-                f"Database contains user with matching sub and name, but user's types do not contain: {types}"
-            )
+            logger.debug(f"Database contains user with matching sub and name, but user's types do not contain: {types}")
             return []
         else:
             return list(set(existing_user_types) & set(types))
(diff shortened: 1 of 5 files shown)


angela-tran:
This PR makes it easier to accomplish #131

eligibility_server/database.py
```diff
@@ -40,12 +40,21 @@ def check_user(self, key: str, user: str, types: str) -> list:
...
             user = self._hash.hash_input(user)
 
-        if (
-            len(types) < 1
-            or key not in self._users
-            or self._users[key][0] != user
-            or len(set(self._users[key][1]) & set(types)) < 1
-        ):
-            return []
+        existing_user = app.User.query.filter_by(user_id=key, key=user).first()
```
thekaveman, line 43: Oof, this is a confusing line 🤔 

`user_id=key, key=user`? I think we should clean up our variables and/or column names.

eligibility_server/database.py
```diff
@@ -40,12 +40,21 @@ def check_user(self, key: str, user: str, types: str) -> list:
...
+            existing_user_types = ast.literal_eval(existing_user.types)
+        else:
+            existing_user_types = None
 
-        return list(set(self._users[key][1]) & set(types))
+        if len(types) < 1:
+            logger.debug("List of types to check was empty.")
+            return []
+        elif existing_user is None:
+            logger.debug(f"Database does not contain requested user with sub, name: {key, user}")
```
thekaveman, line 53: Let's not log the actual user data - this could be sensitive.

eligibility_server/database.py
```diff
@@ -40,12 +40,21 @@ def check_user(self, key: str, user: str, types: str) -> list:
...
+            return []
+        elif len(set(existing_user_types) & set(types)) < 1:
+            logger.debug(f"Database contains user with matching sub and name, but user's types do not contain: {types}")
+            return []
+        else:
+            matching_types = set(existing_user_types) & set(types)
```
thekaveman, line 59: Could this set intersection be created once and stored in a variable for clarity and use?

eligibility_server/database.py
angela-tran: Yeah, I totally agree. I wasn't sure if I should do that in this PR. As you point out though, I'm adding code that uses those variables and column names, so it actually is appropriate in this PR. Will clean up!

angela-tran: Are you ok with the variable/column names being `sub` and `name` to match the API? Or would you prefer `user_id` and `user_name`?

machikoyasuda (approved):

eligibility_server/database.py
```diff
@@ -24,37 +24,38 @@ def __init__(self, hash=False):
...
+        existing_user = app.User.query.filter_by(sub=sub, name=name).first()
         if existing_user:
             existing_user_types = ast.literal_eval(existing_user.types)
         else:
-            existing_user_types = None
+            existing_user_types = []
+
+        matching_types = set(existing_user_types) & set(types)
 
         if len(types) < 1:
```
thekaveman, line 51: Should we move this check all the way to the top, before the DB lookup? No need to do that if there are no types to compare against, right?

angela-tran: Yep, great point

thekaveman (approved):

PR #140, opened 2022-08-31
Refactor: organize application as a package

### LINKED ISSUE #139: Fix setup/teardown logging
After #128 was merged in, the `setup` and `teardown` scripts are no longer showing log statements in the console:

### Additional context
This has to do with logger hierarchy. In #116, we were setting the log level and handler for the [`root`](https://github.com/cal-itp/eligibility-server/pull/116/files#diff-e5991fd559847cf92df779bd8b72643e9b64926355be788735b3c1cfcabca211R29) logger. The intent in #128 was to:
 - introduce a configurable setting for the log level
 - simplify the configuration code (avoid adding/removing redundant handlers)

I thought configuring directly on the `app.logger` instance would be equivalent to our previous setup, but I see now that's not true. 

This bug points out that we should revisit the organization of our scripts and where they get their `Logger`. This might have some overlap with #131.

### Resources
https://docs.python.org/3/howto/logging.html#advanced-logging-tutorial

angela-tran:
Part of #131 

Closes #139 

This PR makes changes to the organization of our code so that 
 - database-related code is grouped together 
 - our application can be imported as a package. 

This should allow us to move towards using the app factory pattern, which in turn should move us towards using a separate database when testing.

Note that the `setup.py` and `teardown.py` scripts for modifying the database were converted into `click` commands to facilitate moving them into a `db` sub-package and since that's the approach shown in [Flask documentation](https://flask.palletsprojects.com/en/2.2.x/tutorial/database/#id3).


docs/getting-started.md
```diff
@@ -83,13 +83,13 @@ In testing the database, you may need to teardown the database and restart a dat
 The teardown script removes all users and drops the database. To tear down the database, run:
 
 ```bash
-python teardown.py
+flask drop-db
```
machikoyasuda, line 86: Line 68 of this doc states: ```Starting the Dev Container will run `bin/init.sh`, which runs `setup.py` and starts the Flask app. The `setup.py` script creates the database and imports and saves users
based on the configured settings.```

It should now read that `bin/init.sh` runs `flask init-db` , which runs `db/setup.py` (or something like this...)

machikoyasuda (changes requested):
Small documentation change.

I ran and tested:
- Rebuilt the container off of `main` branch
- Ran `pip install -r requirements.txt`, `flask init-db`
- Ran and passed all tests
- Opened `http://localhost:5000/healthcheck`, `http://localhost:5000/publickey`
- Tested `flask drop-db`

machikoyasuda:
I believe this PR also closes https://github.com/cal-itp/eligibility-server/issues/139

thekaveman:
👀 Started reading through this tonight, will finish a little later or tomorrow!

setup.py
```diff
@@ -1,74 +1,10 @@
...
-        logger.info("Table created.")
-        import_users()
+from setuptools import setup
+
+setup(
+    name="eligibility_server",
+    packages=["eligibility_server"],
+    include_package_data=True,
+    install_requires=[
+        "flask",
```
thekaveman, line 8: I think here we want to list everything from `requirements.txt`, right?

I've done it like this in the past: https://github.com/compilerla/intake-html-table/blob/main/setup.py#L3

eligibility_server/db/setup.py
```diff
@@ -0,0 +1,76 @@
+import csv
+import json
+
+import click
```
thekaveman, line 4: I didn't see `click` added to the requirements.txt?

eligibility_server/db/teardown.py
```diff
@@ -0,0 +1,24 @@
...
+                click.echo(f"Users to be deleted: {User.query.count()}")
+                User.query.delete()
+                db.session.commit()
+            except Exception as e:
+                click.echo("Failed to query for Users", e)
+
+            db.drop_all()
+            click.echo("Database dropped.")
+        else:
+            click.echo("Database does not exist.")
```
thekaveman, line 24: I would almost suggest combining `teardown.py` into `setup.py`, we still have different commands so it seems unnecessary to separate the modules since we aren't calling them directly anymore (`python setup.py/teardown.py`)

eligibility_server/verify.py
```diff
@@ -8,12 +8,12 @@
 import re
 import time
 
-from flask import abort
+from flask import abort, current_app
```
thekaveman, line 11: I like this approach, feels way cleaner than `app.app.something`

eligibility_server/db/models.py
```diff
@@ -0,0 +1,27 @@
+import logging
+
+from flask_sqlalchemy import SQLAlchemy
+
+logger = logging.getLogger(__name__)
+
+db = SQLAlchemy()
+
+
+def init_app(app):
```
thekaveman, line 10: This feels like a strange module for the initialization of the database. Would it make more sense in `__init__.py`?

eligibility_server/db/setup.py
```diff
@@ -0,0 +1,76 @@
...
+            )
+            for user in data:
+                save_users(db, User, user[0], user[1], user[2])
+    else:
+        click.echo(f"Warning: File format is not supported: {file_format}")
+
+    click.echo(f"Users added: {User.query.count()}")
+
+
+def save_users(db, User, sub: str, name: str, types: str):
```
thekaveman, line 65: I'm confused we pass in the `User` class here, when it is imported globally at the top?

eligibility_server/verify.py
```diff
@@ -26,27 +26,28 @@ def __init__(self):
         self.client_public_key = keypair.get_client_public_key()
         self.server_private_key = keypair.get_server_private_key()
 
-        if app.app.config["INPUT_HASH_ALGO"] != "":
-            hash = Hash(app.app.config["INPUT_HASH_ALGO"])
+        if current_app.config["INPUT_HASH_ALGO"] != "":
```
thekaveman, line 29: Going to create a follow-up issue to hide all these settings keys in a config object that we can import and use its attribute, e.g. `config.hash_algo`

thekaveman (changes requested):
This looks like a good direction overall.

I'm not totally thrilled with the `Database` class (nothing you did in this PR) - it feels weird, we called it that in the past because it literally was the hardcoded database. Now it's just that `check_user` function... which is really doing the eligibility check? Oh well. Not too worried about this for now.

Left a few comments inline, a few areas we need some changes and a few potential cleanups/reorgs.

angela-tran:
Thanks @machikoyasuda and @thekaveman for your thorough reviews! I agree with the things you pointed out and will work through these items.

eligibility_server/db/setup.py
angela-tran: Yeah, that is because `click` gets installed too when you install `flask`.

(https://flask.palletsprojects.com/en/2.2.x/installation/#dependencies)

thekaveman: I had no idea!

eligibility_server/db/setup.py
angela-tran: Oops, I think this is leftover from an intermediate cleanup state I was in. Thanks for catching this! I will clean up.

docs/getting-started.md
angela-tran: Resolved in https://github.com/cal-itp/eligibility-server/pull/140/commits/75dea8ed2b3e9c1ea6d8c306b412baa5333f65e1

eligibility_server/verify.py
angela-tran: I think this can be resolved since #142 has been created

eligibility_server/app.py
```diff
@@ -7,9 +7,9 @@
 from flask_restful import Api
 from flask.logging import default_handler
 
+from . import db
 from .verify import Verify
 from .keypair import get_server_public_key
```
thekaveman, line 12: It might make sense to use the package as the import source now:

```python
from eligibility_server import db
from eligibility_server.verify import Verify
from eligibility_server.keypair import get_server_public_key
```

eligibility_server/db/database.py
```diff
@@ -0,0 +1,59 @@
+import ast
+import logging
+
+from .models import User
```
thekaveman, line 4: Thoughts on the more verbose/explicit:

```python
from eligibility_server.db.models import User
```

angela-tran:
@thekaveman I made some changes based on your feedback. Also, I agree with this:

> I'm not totally thrilled with the `Database` class (nothing you did in this PR) - it feels weird, we called it that in the past because it literally was the hardcoded database. Now it's just that `check_user` function... which is really doing the eligibility check? Oh well. Not too worried about this for now.

Should I try to move the `check_user` logic into the `verify` module as a part of this PR?

eligibility_server/db/models.py
```diff
@@ -1,20 +1,4 @@
...
-
-def init_app(app):
-    db.init_app(app)
-
-    from .setup import init_db_command
-    from .teardown import drop_db_command
-
-    app.cli.add_command(init_db_command)
-    app.cli.add_command(drop_db_command)
+from . import db
```
thekaveman, line 1: Same comment on import style/source

eligibility_server/verify.py
```diff
@@ -13,7 +13,7 @@
 from jwcrypto import jwe, jws, jwt
 
 from . import keypair
-from .db import Database
+from .db.database import Database
 from .hash import Hash
```
thekaveman, line 18: Same comment on import style/source

eligibility_server/db/__init__.py
```diff
@@ -1,58 +1,17 @@
-import ast
 import logging
 
-from .models import User
+from flask_sqlalchemy import SQLAlchemy
 
 logger = logging.getLogger(__name__)
 
 
-class Database:
```
thekaveman: FWIW I wasn't necessarily saying this class should move from here, just that it is a weird name/class.

The import `from eligibility_server.db.database import Database` is a little clunky, but that's fine.

eligibility_server/app.py
```diff
@@ -67,7 +67,7 @@ def internal_server_error(error):
 api = Api(app)
 api.add_resource(Verify, "/verify")
 
-models.init_app(app)
+db.init_app(app)
```
thekaveman, line 70: 👍

eligibility_server/verify.py
```diff
@@ -13,7 +13,7 @@
 from jwcrypto import jwe, jws, jwt
 
 from . import keypair
-from .db import Database
+from .db.database import Database
```
thekaveman, line 16: ^ little clunky (3x database!)

tests/test_database.py
```diff
@@ -4,7 +4,7 @@
 import pytest
 
 
-from eligibility_server.db import Database
+from eligibility_server.db.database import Database
```
thekaveman, line 7: ^ another 3x database

thekaveman (changes requested):
All of my comments below are mostly stylistic and don't _require_ any changes.

But one thing that should be addressed: *We want the Docker app container to install this package!*

Need a line in the `Dockerfile` like:


```docker
RUN pip install -e .
```

Obviously the source and `setup.py` need to the in the image layer for this to work, so may need to adjust how/where `pip install` happens, or just do it a second time if that's preferable.

thekaveman:
> @thekaveman I made some changes based on your feedback. Also, I agree with this:
> 
> > I'm not totally thrilled with the `Database` class (nothing you did in this PR) - it feels weird, we called it that in the past because it literally was the hardcoded database. Now it's just that `check_user` function... which is really doing the eligibility check? Oh well. Not too worried about this for now.
> 
> Should I try to move the `check_user` logic into the `verify` module as a part of this PR?

Didn't see this comment before leaving my review - but yeah that could be a better place for it! Maybe turn `verify` into a sub-package with `api.py` and `somethingelse.py`?

setup.py
```diff
@@ -1,74 +1,36 @@
...
+
+
+CLASSIFIERS = [
+    "License :: OSI Approved :: GNU Affero General Public License v3 or later (AGPLv3+)",
+    "Operating System :: OS Independent",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.7",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
```
thekaveman, line 18: Technically 3.10 could go here too.... seems safe to just say `3` and not specify all the minor versions

setup.py
```diff
@@ -1,74 +1,36 @@
...
+]
+
+setup(
+    name="eligibility-server",
+    description="Server implementation of the Eligibility Verification API",
+    long_description=long_description,
+    long_description_content_type="text/markdown",
+    python_requires=">=3.7",
```
thekaveman, line 26: ^^ Especially in combination with this min version

tests/test_database.py
```diff
@@ -1,33 +0,0 @@
-"""
-Test database class and methods
-"""
```
angela-tran, line 3: I was going to say let's hold off on refactoring these tests until we get a test database going, but on second thought, let me try something

angela-tran: I was able to port these tests over to using an instance of `Verify` instead of `Database`

thekaveman (approved):
Looks great!

* All tests passed locally
* The package is installed (in a python REPL session I could `import eligibility_server`)
* `/healthcheck`, `/publickey`, `/verify` return expected results

machikoyasuda (approved):

PR #123, opened 2022-08-11
Feat: production container

thekaveman:
Closes #111 

This PR changes the base image to use our new [`cal-itp/docker-python-web`](https://github.com/cal-itp/docker-python-web) image hosted on [GitHub Container Registry](https://github.com/cal-itp/docker-python-web/pkgs/container/docker-python-web).

This base image bakes in `nginx` and `gunicorn` - and `bin/start.sh` runs the app through those services.

Developers shouldn't notice a difference running the app locally - either inside the devcontainer in debug mode, or by running the container with `docker compose`.

PR #141, opened 2022-08-31
Chore: specify 8000 as port in launch.json

angela-tran:
@machikoyasuda and I noticed that when launching `eligibility-server` from the VS Code debugger, it was being served at 5000 instead of a dynamic port like it used to.

image

Looks like this started happening when the devcontainer's port was changed from [5000 to 8000](https://github.com/cal-itp/eligibility-server/pull/123/files#diff-3493e6b5ddf34891e572f911db893efd9e46af828e011ea778a7c1eb64763588).

This PR restores the old behavior by specifying in `launch.json` that 8000 should be the port for Flask to serve the app at (instead of the default 5000).

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -16,6 +16,7 @@
             ],
             "env": {
                 "FLASK_DEBUG": "1",
+                "FLASK_RUN_PORT": "8000",
                 "ELIGIBILITY_SERVER_SETTINGS": "../config/sample.py"
             },
         },


machikoyasuda (approved):
image
image
Confirmed localhost:8000 as a the port.

thekaveman (approved):
It doesn't seem to be a problem to add this setting. 

I'll just note the change for serving from port 8000 only matters when running the _application container_ served through `nginx` and `gunicorn` - for the VS Code / devcontainer use-case, the port can be whatever.

PR #146, opened 2022-09-07
Fix: configure root logger

angela-tran:
Related to #139 

This PR makes sure that in addition to configuring Flask's logger (which Flask uses internally), we also configure the root logger. This is so that loggers we create from the standard logging module are configured correctly.

Our code doesn't need to use `app.logger` because I believe there is nothing to be gained from using it over just a normal `logger`, and I think we'd rather use not have to import `current_app` everywhere that needs logging.

## Steps to Reproduce
 - Launch the server with default settings
 - Go to the `/verify` endpoint, e.g. in a browser
 - Look at the logs

## Expected behavior
Since the log level in default settings is `INFO`, you should see at least that level.
Something like:
```
* Tip: There are .env or .flaskenv files present. Do "pip install python-dotenv" to use them.
[2022-09-07 15:27:33,742] WARNING werkzeug:224  * Debugger is active!
[2022-09-07 15:27:33,742] INFO werkzeug:224  * Debugger PIN: 575-180-751
[2022-09-07 15:28:02,697] INFO eligibility_server.keypair:33 Reading client key file: ./keys/client.pub
[2022-09-07 15:28:02,715] INFO eligibility_server.keypair:39 Reading server private key file: ./keys/server.key
[2022-09-07 15:28:02,759] INFO eligibility_server.hash:22 Input hash algorithm set to: sha256
[2022-09-07 15:28:04,762] ERROR eligibility_server.verify:125 Forbidden
[2022-09-07 15:28:04,765] INFO werkzeug:224 172.24.0.1 - - [07/Sep/2022 15:28:04] "GET /verify HTTP/1.1" 403 -
```

## Actual behavior
Not seeing any `INFO` logs, and also not seeing formatted log messages. It is as if the `logger` is not configured.
```
* Tip: There are .env or .flaskenv files present. Do "pip install python-dotenv" to use them.
* Debugger is active!
* Debugger PIN: 575-180-751
Forbidden
172.24.0.1 - - [07/Sep/2022 15:29:46] "GET /verify HTTP/1.1" 403 -
```

### DIFF
--- a/eligibility_server/app.py
+++ b/eligibility_server/app.py
@@ -15,8 +15,15 @@ app = Flask(__name__)
 app.config.from_object("eligibility_server.settings")
 app.config.from_envvar("ELIGIBILITY_SERVER_SETTINGS", silent=True)
 
-app.logger.setLevel(app.config["LOG_LEVEL"])
-default_handler.formatter = logging.Formatter("[%(asctime)s] %(levelname)s %(name)s:%(lineno)s %(message)s")
+log_level = app.config["LOG_LEVEL"]
+format_string = "[%(asctime)s] %(levelname)s %(name)s:%(lineno)s %(message)s"
+
+# configure Flask's logger
+app.logger.setLevel(log_level)
+default_handler.formatter = logging.Formatter(format_string)
+
+# configure root logger
+logging.basicConfig(level=log_level, format=format_string)
 
 
 def TextResponse(content):


thekaveman (commented):
This looks great. 

Any idea why we are getting duplicate entries for the messages that do come from `app.logger`? E.g. `/healthcheck` and `/publickey`:

image

angela-tran:
> This looks great.
> 
> Any idea why we are getting duplicate entries for the messages that do come from `app.logger`? E.g. `/healthcheck` and `/publickey`

Oof, thank you for catching that. Looks like we need to configure the root logger first so that the `app.logger` doesn't add its own handler. 
(If `app.logger` is called first, it triggers [`create_logger()`](https://github.com/pallets/flask/blob/main/src/flask/app.py#L771-L796) which [checks for existing handlers](https://github.com/pallets/flask/blob/main/src/flask/logging.py#L71-L72).)

angela-tran:
After switching the order, I don't see duplicate logging anymore:

>  \* Tip: There are .env or .flaskenv files present. Do "pip install python-dotenv" to use them.
> [2022-09-07 16:46:53,016] WARNING werkzeug:224  * Debugger is active!
> [2022-09-07 16:46:53,017] INFO werkzeug:224  * Debugger PIN: 575-180-751
> **[2022-09-07 16:46:57,722] INFO eligibility_server.app:38 Request healthcheck**
> [2022-09-07 16:46:57,723] INFO werkzeug:224 172.24.0.1 - - [07/Sep/2022 16:46:57] "GET /healthcheck HTTP/1.1" 200 -
> **[2022-09-07 16:47:30,671] INFO eligibility_server.app:44 Request public key**
> [2022-09-07 16:47:30,671] INFO eligibility_server.keypair:45 Reading server public key file: ./keys/server.pub
> [2022-09-07 16:47:30,677] INFO werkzeug:224 172.24.0.1 - - [07/Sep/2022 16:47:30] "GET /publickey HTTP/1.1" 200 -

eligibility_server/app.py
```diff
@@ -15,8 +15,15 @@
 app.config.from_object("eligibility_server.settings")
 app.config.from_envvar("ELIGIBILITY_SERVER_SETTINGS", silent=True)
 
-app.logger.setLevel(app.config["LOG_LEVEL"])
-default_handler.formatter = logging.Formatter("[%(asctime)s] %(levelname)s %(name)s:%(lineno)s %(message)s")
+log_level = app.config["LOG_LEVEL"]
+format_string = "[%(asctime)s] %(levelname)s %(name)s:%(lineno)s %(message)s"
+
+# configure root logger
```
thekaveman, line 21:
```suggestion
# configure root logger first, to prevent duplicate log entries from Flask's logger
```

thekaveman (approved):

PR #144, opened 2022-09-03
Feat: settings/configuration helper

thekaveman:
Closes #142 

~Based on #140, will keep up to date with that branch and wait for it to be merged to `main`.~

Introduces a new `Configuration` class in the `settings.py` module, with `@property` attributes for each of the settings `VARIABLES`. Property names are the `lower_case` version of their setting.

Use it like:

```python
from eligibility_server.settings import Configuration

config = Configuration()

print(config.input_hash_algo)
```

_Note: omitted the `SQLALCHEMY` settings as these are not used directly in the app code._

PR #148, opened 2022-09-07
Refactor: models

### LINKED ISSUE #122: Refactor database class

angela-tran:
Closes #122 

~This will need to be rebased onto #144.~ done

This PR splits the eligibility type out from `User` into its own model called `Eligibility`. There is a many-to-many relationship between these two.

| Table | Before | After |
| ---- | ---- | ---- |
| `user` | image | image |
| `eligibility` | - | image  |
| `user_eligibility` | - | image |

_<details><summary>How to view the tables like that ^^_</summary>

> In a terminal in the devcontainer:
> ```
> ln -s /tmp/test.db db.sqlite
> code db.sqlite
> ```
> (Makes a symlink to the SQLite database file with an extension that VS Code recognizes. Then open it with VS Code.)
> </details>

This PR also makes a few changes (improvements?) to the expected format of CSV data import files:

**`data/server.csv`**
**Before**
```
A1234567;Garcia;['type1']
B2345678;Hernandez;['type2']
C3456789;Smith;[]
D4567890;Jones;['type1', 'type2']
a93ef111829829ea5038398ba6583dc2e1a2f2c309e24aee535f725dbe1f1250;2c3aaefea8267c66822f6edc0e42d9b7384695f9c0407eabda141770aab8901e;['type1']
095feaba889222ee1bc97aaf3bb89c90c21238fb556fb48b5cc54760461ea18ce814d016e70b1929d05190edbec1032e07921228fcb61ad3417aaff0abd9d082;123c86e1f2ac255ba31f1ad742defe23d194269669d2aac0d2572e20e9378e395976f84db305caeba1f91e7996463031d4c49365a7a9f4c7dc404873ad330974;['type1']
```
**After**
```
A1234567;Garcia;type1
B2345678;Hernandez;type2
C3456789;Smith;
D4567890;Jones;type1
D4567890;Jones;type2
a93ef111829829ea5038398ba6583dc2e1a2f2c309e24aee535f725dbe1f1250;2c3aaefea8267c66822f6edc0e42d9b7384695f9c0407eabda141770aab8901e;type1
095feaba889222ee1bc97aaf3bb89c90c21238fb556fb48b5cc54760461ea18ce814d016e70b1929d05190edbec1032e07921228fcb61ad3417aaff0abd9d082;123c86e1f2ac255ba31f1ad742defe23d194269669d2aac0d2572e20e9378e395976f84db305caeba1f91e7996463031d4c49365a7a9f4c7dc404873ad330974;type1
```
or
```
A1234567;Garcia;type1
B2345678;Hernandez;type2
C3456789;Smith;
D4567890;Jones;"type1,type2"
a93ef111829829ea5038398ba6583dc2e1a2f2c309e24aee535f725dbe1f1250;2c3aaefea8267c66822f6edc0e42d9b7384695f9c0407eabda141770aab8901e;type1
095feaba889222ee1bc97aaf3bb89c90c21238fb556fb48b5cc54760461ea18ce814d016e70b1929d05190edbec1032e07921228fcb61ad3417aaff0abd9d082;123c86e1f2ac255ba31f1ad742defe23d194269669d2aac0d2572e20e9378e395976f84db305caeba1f91e7996463031d4c49365a7a9f4c7dc404873ad330974;type1
```

 - Lists of eligibility types are [expected to be separated by `,`](https://github.com/cal-itp/eligibility-server/pull/148/files#diff-7b32e61aaf3f582dd8acd2c253e8b1d7ed91770f99abf04003af72a990f98c8bR59-R60). (We used to expect them as Python-style lists that would be evaluted by `ast.literal_eval`.)
    - If the CSV delimiter is also `,`, then the list should be surrounded by the CSV quote character (specified by `CSV_QUOTCHAR` setting).
 - Removes `ast` module import since we don't need it anymore
 - Adds support for CSVs to contain multiple rows for a given `sub`, `name` pair (i.e. a `User`). I think this better matches what you might see from a SQL query result.

Wrote unit tests for the `init-db` and `drop-db` commands - the `init-db` test is hard-coded to the sample data.


angela-tran:
Although this will have minor merge conflicts with #144, I think this can be reviewed now and spot-checked after the rebase.

eligibility_server/db/setup.py
```diff
@@ -45,34 +45,43 @@ def import_users():
         with open(file_path) as file:
             data = json.load(file)["users"]
             for user in data:
-                save_users(user, data[user][0], str(data[user][1]))
+                save_users(user, data[user][0], data[user][1])
```
machikoyasuda, line 48: I like how this is cleaner than the previous `str()` version.

thekaveman:
We decided today that we're going to assume the CSV file always has the eligibility types in it, and not worry about filling in default types etc.

angela-tran:
Done rebasing on `main` to resolve conflicts with #144 changes. This is ready for review @thekaveman @machikoyasuda

eligibility_server/settings.py
```diff
@@ -41,7 +41,7 @@
 CSV_DELIMITER = ";"
 CSV_NEWLINE = ""
 CSV_QUOTING = 3
-CSV_QUOTECHAR = ""
+CSV_QUOTECHAR = '"'
```
thekaveman, line 44: Would rather this was `"\""` to be consistent with quotes, but I think I needed it in my PR for remote data so I added it there.

eligibility_server/db/setup.py
thekaveman: I do think it's a little funky to pass `user` and then use `user` as a key for `data` - not super obvious what is going on here?

Also this JSON file is... the oldest remnant of the old school approach from our toy. Not that important to maintain / make clearer right now, probably just drop it entirely in the future?

ALSO I've updated this section for the remote data anyway!

thekaveman (commented):
So this is an interesting problem, and a possible solution...

If you run the tests using the VSCode interface, they should all pass. But then immediately run them again without touching anything else - many of the tests in `test_verify.py` will fail.

This is because tests in `db/test_setup.py`  _delete_ the database and it isn't being recreated elsewhere; the verify tests expect it to already exist. This is probably part of what we've been talking/thinking about in #131.

An immediate resolution could be to add a call to `init-db` at the very end of the `test_drop_db_command` test? Simple and kind of a hack, but it would at least make all the tests idempotent (a la #103).

A possible longer term solution could be to abstract that `runner.invoke(args="init-db")` as a fixture (and the `drop-db`) that can be injected/used across tests?

eligibility_server/settings.py
angela-tran: Yeah, I wanted to keep the quotes consistent too, but pre-commit.ci changed it in https://github.com/cal-itp/eligibility-server/pull/148/commits/7ce18bf7aca22efc8d197f6e747773254634c6af.

angela-tran:
> If you run the tests using the VSCode interface, they should all pass. But then immediately run them again without touching anything else - many of the tests in test_verify.py will fail.

Hmm that is interesting. Good catch 🙏

tests/db/test_setup.py
```diff
@@ -0,0 +1,41 @@
...
+@pytest.mark.usefixtures("flask")
+def test_drop_db_command(runner):
+    result = runner.invoke(args="drop-db")
+
+    assert result.exit_code == 0
+
+    inspector = inspect(db.engine)
+    assert inspector.get_table_names() == []
+
+    runner.invoke(args="init-db")
```
thekaveman, line 41:
```suggestion
    # temp fix to ensure database tests are idempotent - later tests need the database to exist
    runner.invoke(args="init-db")
```

thekaveman (approved):

thekaveman:
🎉

PR #151, opened 2022-09-13
Feat: load remote data

### LINKED ISSUE #143: Support data import from remote URLs
Similar to how we can load keys from remote URLs introduced in #135, it might be advantageous to be able to point to a remote URL for the `IMPORT_FILE_PATH` setting.

With a bit of extra logic/settings for auth, this may be how we point to the remote location for e.g. Courtesy Cards data.

thekaveman:
Closes #143.

Simple check if `IMPORT_FILE_PATH` starts with `http`, using `requests.get(path)` to download the file.

* For `json`, file content is loaded to a Python dict
* For `csv`, text content is downloaded to a temp file and read as normal

Also adds debug settings (via `launch.json`) for the database setup/teardown process.

## Testing

1. Update your `IMPORT_FILE_PATH` to point to the raw URL of one of the sample data files, e.g. https://raw.githubusercontent.com/cal-itp/eligibility-server/main/data/server.csv
2. Run `flask drop-db` to drop the existing database, if any
3. In the VS Code _Run and Debug_ tab, select the `Flask: Eligibility Verification Setup` launch item
4. Place breakpoints in the `eligibility_server/db/setup.py` module as appropriate, and hit `F5` to run the `init-db` command

~Sitting on top of #149 for now.~

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,24 +1,39 @@
 {
-    // Use IntelliSense to learn about possible attributes.
-    // Hover to view descriptions of existing attributes.
-    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
-    "version": "0.2.0",
-    "configurations": [
-        {
-            "name": "Flask: Eligibility Verification Server",
-            "type": "python",
-            "request": "launch",
-            "module": "flask",
-            "args": [
-                "run",
-                "--host",
-                "0.0.0.0"
-            ],
-            "env": {
-                "FLASK_DEBUG": "1",
-                "FLASK_RUN_PORT": "8000",
-                "ELIGIBILITY_SERVER_SETTINGS": "../config/sample.py"
-            },
-        },
-    ]
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+  "version": "0.2.0",
+  "configurations": [
+    {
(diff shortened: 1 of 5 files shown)


thekaveman:
I actually have one change to make to this, based on what I'm reading about the `lineterminator` parameter here: https://docs.python.org/3/library/csv.html#csv.Dialect.lineterminator

> The string used to terminate lines produced by the [`writer`](https://docs.python.org/3/library/csv.html#csv.writer). It defaults to `'\r\n'`.
>
> **Note:** The [`reader`](https://docs.python.org/3/library/csv.html#csv.reader) is hard-coded to recognise either `'\r'` or `'\n'` as end-of-line, and ignores _lineterminator_.

thekaveman:
@angela-tran rebased on your database work. Take a look when you get a chance, thanks!

angela-tran (approved):
These changes look good to me and I think can be merged in now.

It'd be nice to have tests for loading of remote JSON and CSV files. (I guess we're also missing a test for loading a local JSON file.) 

We would need to change the value of `IMPORT_FILE_PATH`... maybe that is easier now since we have the `Configuration` object which can be mocked (?). 🤔

thekaveman:
@angela-tran 

> It'd be nice to have tests for loading of remote JSON and CSV files.

Agreed. We would want to fake the request/response to avoid a real network request, similar to how we do in Benefits testing. I think you're on the right path with overriding the `IMPORT_FILE_PATH`, like we've done in [other tests with settings](https://github.com/cal-itp/eligibility-server/blob/main/tests/test_verify.py#L32)

PR #172, opened 2022-11-04
fix: get devcontainer building on Apple M1 hardware

afeld:
Requires a Docker setting change.


machikoyasuda (approved):
Fix confirmed.

I ran `bin/build.sh` in my Terminal, then opened VS Code in F5 mode, and then ran Flask: Eligibility Server to start up the server. Confirmed /healthcheck is working. Ran tests and they passed.

PR #183, opened 2022-11-08
docs(Terraform): add more details around variables

afeld:

PR #185, opened 2022-11-08
chore(Azure): split the Key Vault access for engineers from that for the deployer

afeld:
This should keep the Object ID values from changing when we deploy via the Pipeline vs. from our local machines, as seen [here](https://dev.azure.com/mstransit/courtesy-cards/_build/results?buildId=10&view=logs&j=bc150913-bfac-577d-ab97-4eb22da69064&t=9e090432-7346-5314-b3f7-b310ded2d897&l=67). Builds on https://github.com/cal-itp/eligibility-server/pull/183.

- [x] Set new variables in Pipeline
- [ ] Delete `TF_VAR_VELOCITY_ETL_SERVICE_PRINCIPAL_ID`

### DIFF
--- a/terraform/key_vault.tf
+++ b/terraform/key_vault.tf
@@ -6,15 +6,30 @@ resource "azurerm_key_vault" "main" {
   sku_name            = "standard"
   tenant_id           = data.azurerm_client_config.current.tenant_id
 
-  # we need this to set secrets through the portal
+  # allow engineers to fully manage secrets
   access_policy {
     tenant_id = data.azurerm_client_config.current.tenant_id
-    object_id = data.azurerm_client_config.current.object_id
+    object_id = var.ENGINEERING_GROUP_ID
 
     secret_permissions = [
+      "Backup",
+      "Delete",
       "Get",
-      "Set",
-      "List"
+      "List",
+      "Purge",
+      "Recover",
+      "Restore",
+      "Set"
+    ]
+  }
+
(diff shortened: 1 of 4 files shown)


terraform/key_vault.tf
```diff
@@ -6,15 +6,30 @@ resource "azurerm_key_vault" "main" {
...
+      "Purge",
+      "Recover",
+      "Restore",
+      "Set"
+    ]
+  }
+
+  # allow the Pipeline to read secrets
+  access_policy {
+    tenant_id = data.azurerm_client_config.current.tenant_id
```
angela-tran, line 28: Was curious on what this even is... Read from https://learn.microsoft.com/en-us/azure/active-directory/fundamentals/active-directory-how-to-find-tenant that

> Azure subscriptions have a trust relationship with Azure Active Directory (Azure AD). ...  Each subscription has a tenant ID associated with it

angela-tran (approved):

terraform/key_vault.tf
afeld: Right, it appears to be the ID of the Active Directory domain. Apparently it's public??

https://www.whatismytenantid.com/

angela-tran: Oh interesting. No problem here, I was just putting the link here in case anyone else was curious

angela-tran (approved):

PR #187, opened 2022-11-08
Chore: remove fake delay

angela-tran:


thekaveman (approved):
Yep 👍 this was a silly leftover from the days of Server being a toy inside the Benefits repo, where  we needed some fake latency to test the button animation 🤣

PR #189, opened 2022-11-08
Chore: remove IP restrictions

angela-tran:
Talked with @thekaveman, and these are not a requirement for the initial release. We will revisit this in the next release.

### DIFF
--- a/terraform/app_service.tf
+++ b/terraform/app_service.tf
@@ -19,6 +19,8 @@ resource "azurerm_linux_web_app" "main" {
   https_only          = true
 
   site_config {
+    # required to remove the restrictions
+    ip_restriction         = []
     ftps_state             = "Disabled"
     vnet_route_all_enabled = true
     application_stack {
--- a/terraform/app_service.tf
+++ b/terraform/app_service.tf
@@ -19,13 +19,7 @@ resource "azurerm_linux_web_app" "main" {
   https_only          = true
 
   site_config {
-    ftps_state = "Disabled"
-    dynamic "ip_restriction" {
-      for_each = var.IP_ADDRESS_WHITELIST
-      content {
-        ip_address = ip_restriction.value
-      }
-    }
+    ftps_state             = "Disabled"
     vnet_route_all_enabled = true
     application_stack {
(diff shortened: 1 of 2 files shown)


afeld:
Why don't we just clear the variable? No code change needed.

afeld (dismissed):

terraform/app_service.tf
```diff
@@ -19,13 +19,7 @@ resource "azurerm_linux_web_app" "main" {
...
 
   site_config {
-    ftps_state = "Disabled"
-    dynamic "ip_restriction" {
-      for_each = var.IP_ADDRESS_WHITELIST
-      content {
-        ip_address = ip_restriction.value
-      }
-    }
+    ftps_state             = "Disabled"
```
afeld, line 22: In that case:

```suggestion
    # required to remove the restrictions
    ip_restriction         = []
    ftps_state             = "Disabled"
```

afeld: Wait, nevermind, you already said this.

angela-tran: Actually I agree with you - I think being explicit like this is better. Also, it turns out the pipeline put back the IP restriction in a [recent build](https://dev.azure.com/mstransit/courtesy-cards/_build/results?buildId=88&view=logs&j=bc150913-bfac-577d-ab97-4eb22da69064&t=9e090432-7346-5314-b3f7-b310ded2d897) 🙃

afeld:
I merged another branch and the manual change got undone, so to get the dev site up again, I changed the variable in the pipeline, manually triggered a deploy of `dev`, and manually removed the IP restriction in the Portal. Wanted to make sure UAT wasn't blocked, hope that was ok.

I don't think it's necessary to merge this since we intend to undo it soon, but your call.

afeld (approved):
Your call about whether to merge or not. Good that it's expliclt, but shouldn't change anything.

angela-tran:
Since the variable has been cleared, we don't expect to accidentally put the restriction back. Going to just close this PR and leave the harmless dynamic block in since we'll come back to it in the next release.

PR #201, opened 2022-11-15
fix: enable strict transport security

afeld:
Minor security improvement.

### DIFF
--- a/eligibility_server/app.py
+++ b/eligibility_server/app.py
@@ -75,6 +75,12 @@ def internal_server_error(error):
     return jsonify(error=f"{error.code} {error.name}: Internal server error"), 500
 
 
+@app.after_request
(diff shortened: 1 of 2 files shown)


eligibility_server/app.py
```diff
@@ -75,6 +75,12 @@ def internal_server_error(error):
     return jsonify(error=f"{error.code} {error.name}: Internal server error"), 500
 
 
+@app.after_request
+def apply_caching(response):
```
angela-tran, line 79: The name `apply_caching` seems different from what this function is doing. Maybe `add_headers`? or just `after_request`?

afeld: D'oh, good catch! Sloppy copy-and-paste job on my part.

PR #205, opened 2022-11-18
faster deploy

afeld:
Removes the redundant `plan`. Also made a couple other small tweaks.

### DIFF
--- a/terraform/pipeline/azure-pipelines.yml
+++ b/terraform/pipeline/azure-pipelines.yml
@@ -11,11 +11,11 @@ trigger:
     include:
       - terraform/*
 stages:
-  - stage: plan
+  - stage: terraform
     pool:
       vmImage: ubuntu-latest
     jobs:
-      - job: plan
+      - job: terraform
         variables:
           - name: OTHER_SOURCE
             value: $[variables['System.PullRequest.SourceBranch']]
--- a/terraform/pipeline/azure-pipelines.yml
+++ b/terraform/pipeline/azure-pipelines.yml
@@ -70,7 +70,7 @@ stages:
               command: plan
               # wait for lock to be released, in case being used by another pipeline run
               # https://discuss.hashicorp.com/t/terraform-plan-wait-for-lock-to-be-released/6870/2
-              commandOptions: -input=false -lock-timeout=2m
+              commandOptions: -input=false -lock-timeout=5m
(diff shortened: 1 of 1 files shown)


terraform/pipeline/azure-pipelines.yml
```diff
@@ -70,17 +70,19 @@ stages:
               command: plan
               # wait for lock to be released, in case being used by another pipeline run
               # https://discuss.hashicorp.com/t/terraform-plan-wait-for-lock-to-be-released/6870/2
-              commandOptions: -input=false -lock-timeout=2m
+              commandOptions: -input=false -lock-timeout=5m
               workingDirectory: "$(System.DefaultWorkingDirectory)/terraform"
               # service connection
               environmentServiceNameAzureRM: deployer
+            # the plan is done as part of the apply (below), so don't bother doing it twice
+            condition: notIn(variables['Build.SourceBranchName'], 'dev', 'test', 'prod')
```
afeld, line 78: The important part.

angela-tran (approved):

afeld:
[Worked!](https://dev.azure.com/mstransit/courtesy-cards/_build/results?buildId=118&view=logs&j=e2b2897c-f736-5615-5c9c-c20ddcd4aa73&t=64d2c667-b30c-5507-42a1-d7f546f77724)

thekaveman:
@afeld would you mind explaining what this means (from the link): 

```console
Evaluating: notIn(variables['Build.SourceBranchName'], 'dev', 'test', 'prod')
Expanded: notIn('dev', 'dev', 'test', 'prod')
Result: False
```

Is the first parameter to `notIn` the thing that is being checked against the rest of the parameters? Or, why is `dev` in there twice? And why is this `False`?

afeld:
> Is the first parameter to `notIn` the thing that is being checked against the rest of the parameters? 

[Exactly.](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#notin)

> why is `dev` in there twice?

It's showing that `variables['Build.SourceBranchName']` evaluates ("expands") to `'dev'`.

> why is this `False`?

The equivalent Python expression would be:

```python
if variables['Build.SourceBranchName'] not in ['dev', 'test', 'prod']:
    # run this task
```

`'dev'` _is_ in that list, so the task is skipped.

thekaveman:
Got it, thanks!

PR #206, opened 2022-11-18
Chore(test): remove deprecated algorithm

angela-tran:
Our pytest check has been failing (e.g. the [one](https://github.com/cal-itp/eligibility-server/actions/runs/3499759820/jobs/5861669208) for #205)

@thekaveman found https://github.com/python/cpython/issues/91257, which says:

> The hashlib.algorithms_available set includes algorithms like ripemd160 and whirlpool, those algorithms are not usable unless openssl legacy provider is loaded. Since it's not loaded, and the hashlib module won't load it, any attempt to use them fails.

We'll remove `whirlpool` from our test cases.


thekaveman (approved):
See https://github.com/python/cpython/issues/91257

We aren't using this algorithm in production and so OK to remove from the tests.

Issue #213, opened 2022-11-22
Decide how to configure environment-specific pipeline variables


angela-tran:
There are cases where we would need environment-specific pipeline variables. For example, the `dev` environment might need a different list of allowed IP addresses as opposed to the ones for `test` and `prod`.

Research needs to be done on how this can be accomplished. Should we just define variables with the environment name appended? e.g. `IP_ADDRESS_WHITELIST_DEV`

angela-tran:
Some ideas here: https://stackoverflow.com/questions/45587553/terraform-environment-specific-variables

afeld:
Another option would be to create separate pipelines per environment, potentially using a [template](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/templates?view=azure-devops) across them. The solution you found of [taking the values in through a `map`](https://stackoverflow.com/a/49621326/358804) is simpler though.

thekaveman:
We have something working right now for handling the IP address allowlists, and don't have additional use-cases for this change.

PR #222, opened 2022-11-30
Feat(Terraform): front door

### LINKED ISSUE #208: Restrict server to respond only to a list of IP addresses
### IP restrictions
This could be done via `ip_restrictions` on the App Service. See #181 for the work done so far on this.

Theoretically, we should be able to use the values from [outbound_ip_address_list](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/linux_web_app#outbound_ip_address_list) for the Benefits app service, but this was not working so we ended up [removing the variable value](https://github.com/cal-itp/eligibility-server/issues/189) for the initial launch.

### Front door
Do some research to see if using a [front door](https://azure.microsoft.com/en-us/products/frontdoor/) on eligibility-server would simplify the networking details for IP restriction. Also see if the Front door provides other security measures that we'd like to have.

angela-tran:
Closes #208 

This PR sets up a Front Door with a WAF policy to allow only a certain list of IP addresses.

Note there is some overlap with #213 since the whitelist is different per-environment.

### Before merging:
 - [x] Clean up manually created Front Door resources that were used to get a working configuration
 - [x] Set `TF_VAR_IP_ADDRESS_WHITELIST_DEV` variable in Azure pipeline
 
### After merging:
 - [ ] Update Benefits dev data migration to use front-door hostname in `EligibilityVerifier` API URL

### Some reference material
- [Configure an IP restriction rule with a Web Application Firewall for Azure Front Door](https://learn.microsoft.com/en-us/azure/web-application-firewall/afds/waf-front-door-configure-ip-restriction)
- [Secure traffic to Azure Front Door origins](https://learn.microsoft.com/en-us/azure/frontdoor/origin-security?tabs=app-service-functions&pivots=front-door-standard-premium)
- [Best practices for Front Door](https://learn.microsoft.com/en-us/azure/frontdoor/best-practices)
- [Best practices for Web Application Firewall (WAF) on Azure Front Door](https://learn.microsoft.com/en-us/azure/web-application-firewall/afds/waf-front-door-best-practices)

### DIFF
--- a/terraform/app_service.tf
+++ b/terraform/app_service.tf
@@ -23,6 +23,24 @@ resource "azurerm_linux_web_app" "main" {
     http2_enabled = true
 
     vnet_route_all_enabled = true
+
+    ip_restriction {
+      name        = "Front Door"
+      priority    = 100
+      action      = "Allow"
+      service_tag = "AzureFrontDoor.Backend"
+      headers {
+        x_azure_fdid = [azurerm_cdn_frontdoor_profile.main.resource_guid]
+      }
+    }
+
+    ip_restriction {
+      name        = "Availability Test"
+      priority    = 200
+      action      = "Allow"
+      service_tag = "ApplicationInsightsAvailability"
+    }
+
     application_stack {
       docker_image     = "ghcr.io/cal-itp/eligibility-server"
       docker_image_tag = local.env_name
--- a/terraform/app_service.tf
+++ b/terraform/app_service.tf
@@ -22,12 +22,6 @@ resource "azurerm_linux_web_app" "main" {
     ftps_state    = "Disabled"
     http2_enabled = true
 
-    dynamic "ip_restriction" {
-      for_each = var.IP_ADDRESS_WHITELIST
-      content {
-        ip_address = ip_restriction.value
-      }
-    }
     vnet_route_all_enabled = true
     application_stack {
       docker_image     = "ghcr.io/cal-itp/eligibility-server"
--- a/terraform/front_door.tf
+++ b/terraform/front_door.tf
@@ -82,8 +82,8 @@ resource "azurerm_cdn_frontdoor_firewall_policy" "main" {
 
     match_condition {
       match_variable = "RequestUri"
-      operator       = "EndsWith"
-      match_values   = ["/healthcheck"]
+      operator       = "Equals"
+      match_values   = ["https://${azurerm_cdn_frontdoor_endpoint.main.host_name}:443/healthcheck"]
     }
   }
 
--- a/terraform/front_door.tf
+++ b/terraform/front_door.tf
@@ -72,10 +72,24 @@ resource "azurerm_cdn_frontdoor_firewall_policy" "main" {
   custom_block_response_body        = base64encode("Blocked")
 
   custom_rule {
-    name     = "iprestriction${local.env_name}"
+    name     = "healthcheck"
     enabled  = true
     type     = "MatchRule"
     priority = 1
+    action   = "Allow"
+
+    match_condition {
+      match_variable = "RequestUri"
+      operator       = "EndsWith"
+      match_values   = ["/healthcheck"]
+    }
+  }
+
+  custom_rule {
+    name     = "iprestriction${local.env_name}"
+    enabled  = true
+    type     = "MatchRule"
+    priority = 2
     action   = "Block"
 
     match_condition {
--- a/terraform/front_door.tf
+++ b/terraform/front_door.tf
@@ -18,6 +18,8 @@ resource "azurerm_cdn_frontdoor_origin_group" "main" {
   name                     = local.front_door_name
   cdn_frontdoor_profile_id = azurerm_cdn_frontdoor_profile.main.id
 
+  # this block is required, and it's empty because we are fine with using the default values
+  # https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/cdn_frontdoor_origin_group#load_balancing
   load_balancing {}
 }
 
--- a/terraform/front_door.tf
+++ b/terraform/front_door.tf
@@ -17,7 +17,6 @@ resource "azurerm_cdn_frontdoor_endpoint" "main" {
 resource "azurerm_cdn_frontdoor_origin_group" "main" {
   name                     = local.front_door_name
   cdn_frontdoor_profile_id = azurerm_cdn_frontdoor_profile.main.id
-  session_affinity_enabled = true
 
   load_balancing {}
 }
--- a/terraform/front_door.tf
+++ b/terraform/front_door.tf
@@ -69,7 +69,7 @@ resource "azurerm_cdn_frontdoor_firewall_policy" "main" {
   enabled                           = true
   mode                              = "Prevention"
   custom_block_response_status_code = 403
-  custom_block_response_body        = base64encode("Blocked")
+  custom_block_response_body        = base64encode("Forbidden")
 
   custom_rule {
     name     = "healthcheck"
--- a/terraform/front_door.tf
+++ b/terraform/front_door.tf
@@ -27,11 +27,8 @@ resource "azurerm_cdn_frontdoor_origin" "main" {
 
   enabled                        = true
   host_name                      = azurerm_linux_web_app.main.default_hostname
-  http_port                      = 80
-  https_port                     = 443
   origin_host_header             = azurerm_linux_web_app.main.default_hostname
   certificate_name_check_enabled = true
-  priority                       = 1
   weight                         = 1000
 }
 
--- a/terraform/front_door.tf
+++ b/terraform/front_door.tf
@@ -1,0 +1,51 @@
+locals {
+  front_door_origin_group_name = azurerm_linux_web_app.main.name
+  front_door_origin_name       = azurerm_linux_web_app.main.name
+  front_door_route_name        = azurerm_linux_web_app.main.name
+}
+
+resource "azurerm_cdn_frontdoor_profile" "main" {
+  name                = "mst-courtesy-cards-${local.env_name}"
+  resource_group_name = data.azurerm_resource_group.main.name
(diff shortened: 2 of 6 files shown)


terraform/variables.tf
```diff
@@ -16,7 +16,19 @@ variable "VELOCITY_ETL_APP_OBJECT_ID" {
...
+  default     = []
+}
+
+variable "IP_ADDRESS_WHITELIST_TEST" {
+  description = "List of IP addresses allowed to connect to the app service, in CIDR notation: https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/linux_web_app#ip_address. By default, all IP addresses are allowed."
+  type        = list(string)
+  default     = []
+}
+
+variable "IP_ADDRESS_WHITELIST_PROD" {
```
angela-tran, line 31: Any ideas for a better way to do this?

angela-tran: We could go with this for now and also leave #213 open for this to be refactored.

thekaveman: I think it's fine to leave a refactor for #213, this seems like a good approach to get started.

terraform/app_service.tf
```diff
@@ -22,13 +22,25 @@ resource "azurerm_linux_web_app" "main" {
...
       }
     }
-    vnet_route_all_enabled = true
+
+    ip_restriction {
+      name        = "Availability Test"
+      priority    = 200
+      action      = "Allow"
+      service_tag = "ApplicationInsightsAvailability"
+    }
```
afeld, line 42: Mind adding a note about this being required to the `uptime` module README? Also a note in the README in this directory about access being restricted, so future folks aren't confused about getting 403s?

terraform/front_door.tf
```diff
@@ -0,0 +1,92 @@
...
+resource "azurerm_cdn_frontdoor_endpoint" "main" {
+  # used in the front door URL
+  name                     = "mst-courtesy-cards-eligibility-server-${local.env_name}"
+  cdn_frontdoor_profile_id = azurerm_cdn_frontdoor_profile.main.id
+}
+
+resource "azurerm_cdn_frontdoor_origin_group" "main" {
+  name                     = local.front_door_name
+  cdn_frontdoor_profile_id = azurerm_cdn_frontdoor_profile.main.id
+  session_affinity_enabled = true
```
afeld, line 20: Minor: Since (I don't think) we're using cookies, not sure this is necessary.

```suggestion
```

terraform/front_door.tf
```diff
@@ -0,0 +1,92 @@
...
+  name                          = local.front_door_name
+  cdn_frontdoor_origin_group_id = azurerm_cdn_frontdoor_origin_group.main.id
+
+  enabled                        = true
+  host_name                      = azurerm_linux_web_app.main.default_hostname
+  http_port                      = 80
+  https_port                     = 443
+  origin_host_header             = azurerm_linux_web_app.main.default_hostname
+  certificate_name_check_enabled = true
+  priority                       = 1
```
afeld, line 35: Minor: The numbers above are all same as [the defaults](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/cdn_frontdoor_origin#arguments-reference), so could be removed for brevity. Fine either way.

terraform/front_door.tf
```diff
@@ -0,0 +1,92 @@
...
+    action   = "Block"
+
+    match_condition {
+      match_variable     = "SocketAddr"
+      operator           = "Contains"
+      negation_condition = true
+      match_values       = local.is_prod ? var.IP_ADDRESS_WHITELIST_PROD : local.is_test ? var.IP_ADDRESS_WHITELIST_TEST : var.IP_ADDRESS_WHITELIST_DEV
+    }
+  }
+}
```
afeld, line 92: Thoughts about allowing access to `/healthcheck` from any IP? Could be useful for troubleshooting, but maybe not worth the extra complexity / attack surface.

afeld (changes requested):
Nice work! Just one blocker; the rest are optional. Questions:

- Thoughts about adding a bit of documentation around why we're using Front Door? Or sufficient to have [this comment](https://github.com/cal-itp/eligibility-server/issues/208#issuecomment-1332859410)?
- We are allowing the availability tests through, which is good… What do we think about having a smoke test to ensure other requests are _not_ allowed? Could do this through [a GitHub Action](https://github.com/marketplace/actions/http-status) or [Azure Pipeline task](https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/invoke-rest-api-v1?view=azure-pipelines&tabs=yaml).

terraform/front_door.tf
thekaveman: Agreed this could be useful, not sure how hard it would be to add here

terraform/front_door.tf
angela-tran: Done in f1bd673da090fbd303eb6636ff1899f81501114e

terraform/front_door.tf
angela-tran: Done in 0c160c72a731e6565dc64b3e86ac2aba4354040f

terraform/front_door.tf
angela-tran: I was able to get a working configuration the `dev` environment. Defined the Terraform equivalent in 5a28a9bcece6cb5084dbbde702768bd1c030a774

angela-tran:
> Thoughts about adding a bit of documentation around why we're using Front Door? Or sufficient to have https://github.com/cal-itp/eligibility-server/issues/208#issuecomment-1332859410?

I added a section to the README in 5e6f74e68084f4b5c47019325b489525876c8796

angela-tran:
> We are allowing the availability tests through, which is good… What do we think about having a smoke test to ensure other requests are not allowed? Could do this through [a GitHub Action](https://github.com/marketplace/actions/http-status) or [Azure Pipeline task](https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/invoke-rest-api-v1?view=azure-pipelines&tabs=yaml).

Any objection to creating a new issue for this?

terraform/front_door.tf
```diff
@@ -0,0 +1,102 @@
...
+  # used in the front door URL
+  name                     = "mst-courtesy-cards-eligibility-server-${local.env_name}"
+  cdn_frontdoor_profile_id = azurerm_cdn_frontdoor_profile.main.id
+}
+
+resource "azurerm_cdn_frontdoor_origin_group" "main" {
+  name                     = local.front_door_name
+  cdn_frontdoor_profile_id = azurerm_cdn_frontdoor_profile.main.id
+
+  load_balancing {}
```
thekaveman, line 21: why the empty `load_balancing` section here? (maybe leave a small comment)

terraform/front_door.tf
```diff
@@ -0,0 +1,102 @@
...
+}
+
+resource "azurerm_cdn_frontdoor_firewall_policy" "main" {
+  name                              = "${local.env_name}waf"
+  resource_group_name               = data.azurerm_resource_group.main.name
+  sku_name                          = azurerm_cdn_frontdoor_profile.main.sku_name
+  enabled                           = true
+  mode                              = "Prevention"
+  custom_block_response_status_code = 403
+  custom_block_response_body        = base64encode("Blocked")
```
thekaveman, line 72: The term most often associated with `403` is _Forbidden_: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403

terraform/front_door.tf
```diff
@@ -0,0 +1,102 @@
...
+  custom_rule {
+    name     = "healthcheck"
+    enabled  = true
+    type     = "MatchRule"
+    priority = 1
+    action   = "Allow"
+
+    match_condition {
+      match_variable = "RequestUri"
+      operator       = "EndsWith"
```
thekaveman, line 83: This mean any request that ends with `/healthcheck` will go through, right?

Should we do an exact match here to shore up this one exception and make sure it is the only one? I think `Equals` is the `operator` for that: https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/cdn_frontdoor_firewall_policy#operator

terraform/front_door.tf
```diff
@@ -0,0 +1,102 @@
...
+  custom_rule {
+    name     = "iprestriction${local.env_name}"
+    enabled  = true
+    type     = "MatchRule"
+    priority = 2
+    action   = "Block"
+
+    match_condition {
+      match_variable     = "SocketAddr"
+      operator           = "Contains"
```
thekaveman, line 97: Same question on this `operator`: is `Contains` too loose here? Can't we expect an exact `Equals` match for the IP address and not allow anything else?

thekaveman (commented):
Left some questions - not necessarily required changes, but hoping for some additional clarity ahead of Approval.

afeld (commented):
We currently have the uptime check accessing the App Service directly, not through Front Door. Thinking we should change it to the latter to make it end-to-end. This would remove the need for allowing those requests to the App Service, though not a big deal to leave them.

thekaveman:
> We currently have the uptime check accessing the App Service directly, not through Front Door. Thinking we should change it to the latter to make it end-to-end. This would remove the need for allowing those requests to the App Service, though not a big deal to leave them.

Seems like a good follow-up! #223

terraform/front_door.tf
angela-tran: Added a comment in 8ed904ee58246672abc6deeaf262ba237096e7b0

terraform/front_door.tf
angela-tran: Done in f9b039d31aef260bdc9fce1633817a2f43a0193d

thekaveman (approved):
Looks like a great start, thank you @angela-tran! We can worry about deployment when you're back 😄

angela-tran:
@afeld let me know if there's anything else to change here. This needs your approval before I can merge. Thanks!

afeld (approved):

Issue #220, opened 2022-11-28
resolve nightly downtime alerts


afeld:
[Discussion thread.](https://cal-itp.slack.com/archives/C022HHSEE3F/p1669650929276489) In short, we have intentional nightly downtime, and this is leading to downtime alerts. We should either decide that we're ok with that downtime and adjust the monitoring to expect that (eliminating what is effectively a false positive), or make it so that downtime isn't necessary.

PR #241, opened 2023-02-16
fix(Azure): suppress downtime alerts when data is being loaded

afeld:
Fixes https://github.com/cal-itp/eligibility-server/issues/220.

Uses an [alert processing rule](https://learn.microsoft.com/en-us/azure/azure-monitor/alerts/alerts-processing-rules) to disable notifications during that specific window of time every night.

### DIFF
--- a/terraform/suppress.arm.json
+++ b/terraform/suppress.arm.json
@@ -1,0 +1,37 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "metricAlertID": {
+      "type": "string"
+    }
+  },
+  "resources": [
+    {
+      "type": "Microsoft.AlertsManagement/actionRules",
+      "apiVersion": "2021-08-08",
+      "name": "suppress-nightly-downtime",
+      "location": "Global",
+      "properties": {
+        "description": "Suppresses alerts during the scheduled nightly downtime",
+        "scopes": ["[parameters('metricAlertID')]"],
+        "actions": [
+          {
+            "actionType": "RemoveAllActionGroups"
+          }
+        ],
+        "schedule": {
+          "timeZone": "Eastern Standard Time",
+          "recurrences": [
+            {
+              "recurrenceType": "Daily",
+              "startTime": "06:00:00",
+              "endTime": "06:10:00"
+            }
+          ]
+        },
+        "enabled": true
+      }
+    }
+  ]
+}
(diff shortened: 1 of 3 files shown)


terraform/uptime.tf
```diff
@@ -8,3 +8,27 @@ module "healthcheck" {
...
+  }
+
+  suppression {
+    recurrence_type = "Daily"
+
+    schedule {
+      start_date_utc = "2020-01-01T00:00:00Z"
+      end_date_utc   = "2050-01-01T00:00:00Z"
+      start_time_utc = "11:00AM"
+      end_time_utc   = "11:10AM"
```
afeld, line 31: Unfortunately, this isn't supported yet: https://github.com/hashicorp/terraform-provider-azurerm/issues/16726 I unfortunately think that means this pull request is blocked.

terraform/uptime.tf
```diff
@@ -8,3 +8,27 @@ module "healthcheck" {
   name = "mst-courtesy-cards-eligibility-server-${local.env_name}-healthcheck"
   url  = "https://${azurerm_linux_web_app.main.default_hostname}/healthcheck"
 }
+
+# ignore when app restarts as data is being reloaded
```
afeld, line 12: Is this only configured on the Velocity side, or somewhere on the Eligibility Server side as well?

terraform/uptime.tf
afeld: Found a workaround.

afeld:
Unfortunately we won't really know if this works until we merge it and see if an alert comes through overnight 🤷

terraform/uptime.tf
```diff
@@ -8,3 +8,20 @@ module "healthcheck" {
...
+# https://learn.microsoft.com/en-us/azure/azure-monitor/alerts/alerts-processing-rules
+# the Terraform resource doesn't support time windows, so need to drop down to an ARM template instead
+# https://github.com/hashicorp/terraform-provider-azurerm/issues/16726
```
afeld, line 15: See this note re: the JSON above.

terraform/suppress.arm.json
```diff
@@ -0,0 +1,37 @@
...
+      "properties": {
+        "description": "Suppresses alerts during the scheduled nightly downtime",
+        "scopes": ["[parameters('metricAlertID')]"],
+        "actions": [
+          {
+            "actionType": "RemoveAllActionGroups"
+          }
+        ],
+        "schedule": {
+          "timeZone": "Eastern Standard Time",
```
afeld, line 24: Created based on grabbing JSON from the Portal, hence it being in my time zone. Didn't bother to change.

angela-tran (approved):
Thanks for the walkthrough @afeld

afeld:
Confirmed we're not longer getting the alerts in dev.

Screenshot 2023-02-27 at 2 47 57 PM

Issue #249, opened 2023-03-01
Make ignoring of downtime more robust


afeld:
Follow-up to https://github.com/cal-itp/eligibility-server/issues/220.

Right now, there seem to be several minutes of downtime on each deploy/restart of the app. While not ideal, this is expected, and thus we probably don't want an alert when that happens. [Alert rules](https://learn.microsoft.com/en-us/azure/azure-monitor/alerts/alerts-create-new-alert-rule?tabs=metric) support conditions, so we could theoretically modify the Monitor to only alert for downtimes of over a few minutes. That would then mean the scheduled alert suppression added in https://github.com/cal-itp/eligibility-server/pull/241 could be removed.

thekaveman:
This hasn't been an ongoing concern that we'd like to address right now.

PR #251, opened 2023-04-10
Fix: expand suppression window

angela-tran:
#248 expanded the `endTime` of the suppression window.

We are still seeing alerts though, and the pattern seems to be that the "Start Time" of the 5-minute evaluation window is _before_ 5:00AM CT.

image

This PR simply expands the `startTime`, and I believe we shouldn't see any more alerts from prod. #249 is a more robust approach, but I figured we could get this smaller fix in for now.

### DIFF
--- a/terraform/suppress.arm.json
+++ b/terraform/suppress.arm.json
@@ -25,7 +25,7 @@
           "recurrences": [
             {
               "recurrenceType": "Daily",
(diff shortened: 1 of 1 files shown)


thekaveman (approved):
👍 good with a simple fix for now

PR #266, opened 2023-06-22
Fix: increase container startup time

### LINKED ISSUE #265: Add WEBSITES_CONTAINER_START_TIME_LIMIT to Terraform config
See Slack threads:

* https://cal-itp.slack.com/archives/C022HHSEE3F/p1687410023429539 (new issue)
* https://cal-itp.slack.com/archives/C022HHSEE3F/p1671492346475199?thread_ts=1671487252.396439&cid=C022HHSEE3F (prior discussion)

TLDR; the app can take too long to restart with new data, causing Azure to restart the whole service on the presumption that it is hung. This setting will increase the time allowed for the app to restart during data reload.

See also this SO post: https://stackoverflow.com/questions/51103897/increase-startup-timeout-for-azure-webapp-for-containers

angela-tran:
Fixes #265 

Documentation on the setting: https://learn.microsoft.com/en-us/azure/app-service/reference-app-settings?tabs=kudu%2Cdotnet#custom-containers

### DIFF
--- a/terraform/app_service.tf
+++ b/terraform/app_service.tf
@@ -53,6 +53,7 @@ resource "azurerm_linux_web_app" "main" {
     # this prevents the filesystem from being obscured by a mount
     "WEBSITES_ENABLE_APP_SERVICE_STORAGE" = "false"
     "WEBSITES_PORT"                       = "8000"
+    "WEBSITES_CONTAINER_START_TIME_LIMIT" = "1800"
   }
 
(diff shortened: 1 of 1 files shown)


thekaveman (approved):

thekaveman:
This value is expressed in seconds, and this PR sets it to the maximum allowed value of 1800, or 30 minutes 👍

PR #285, opened 2023-08-07
Investigating infra pipeline logic

angela-tran:
I'm opening this to see what the Azure pipeline does when opening a PR from `dev` into `test`. 

Based on the code, it seems like it will run `terraform apply` on the "test" workspace even though all we did was open a PR. This implies the `terraform apply` that runs on merging the PR would be a no-op. (I'm not actually going to merge this PR. Just want to look at the results of opening the PR.)


angela-tran:
The infra pipeline ran `terraform plan` on the test workspace, so it works as one would think it should - i.e. no bug! 🙌 



   


The code (i.e. [condition for running `terraform plan`]()) is just a little misleading because of the nuance of `Build.SourceBranchName`. I was able to find [#177](https://github.com/cal-itp/eligibility-server/pull/177/files) which helped me realize that `Build.SourceBranchName` is only populated on `IndividualCI` runs (i.e. merges).

PR #310, opened 2023-08-31
Feat: infra pipeline tag trigger

### LINKED ISSUE #284: Add support for running infra pipeline when Git tag is pushed with specific name
- [x] Decide on naming pattern for Git tags for test and prod environments

### Do in isolated testing environment
- [x] Update `azure-pipeline.yml` to be triggered by Git tag for specific naming patterns
- [x] Update logic that decides when to use `test` and prod/`default` Terraform workspaces (logic is in `workspace.py`)
- [x] Update `azure-pipeline.yml` to require approval for stage that runs `terraform apply` for test/prod deployment
   - [x] Split up pipeline into multiple stages to allow for this
   - [x] Get `terraform plan` stage to save a `tfplan` for the `apply` command to use
- [x] ~Make sure the stage that runs `terraform plan` always runs for test/prod deployment~ already works as written
- [x] Make sure the stage that runs `terraform apply` runs for tag-based test/prod deployment
- [x] Code cleanup: make the two stages share variables

Done in #307 

### Make changes to real pipeline
- [x] Update `azure-pipeline.yml` to be triggered by Git tag for specific naming patterns
- [x] Update logic that decides when to use `test` and prod/`default` Terraform workspaces (logic is in `workspace.py`)
- [x] Update `azure-pipeline.yml` to require approval for stage that runs `terraform apply` for test/prod deployment
- [x] ~Make sure the stage that runs `terraform plan` always runs for test/prod deployment~ already works as written
- [x] Make sure the stage that runs `terraform apply` runs for tag-based test/prod deployment

Done in #310

angela-tran:
Closes #284 

This PR makes it possible for our existing pipeline in MST's DevOps to be triggered by pushing a git tag.

The only thing not defined in code here is the Environment which needs to be created manually and configured manually. 
 - [x] Create environment in MST's DevOps

### DIFF
--- a/terraform/pipeline/azure-pipelines.yml
+++ b/terraform/pipeline/azure-pipelines.yml
@@ -1,6 +1,4 @@
 trigger:
-  # automatically runs on pull requests
-  # https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#pr-triggers
   branches:
     include:
       - dev
@@ -11,6 +9,14 @@ trigger:
       - 20??.??.?*-rc?*
       - 20??.??.?*
   # only run for changes to Terraform files
+  paths:
+    include:
+      - terraform/*
+
+pr:
+  branches:
+    include:
+      - "*"
   paths:
     include:
       - terraform/*
--- a/terraform/pipeline/azure-pipelines.yml
+++ b/terraform/pipeline/azure-pipelines.yml
@@ -6,10 +6,15 @@ trigger:
       - dev
       - test
       - prod
+  tags:
+    include:
+      - 20??.??.?*-rc?*
+      - 20??.??.?*
   # only run for changes to Terraform files
   paths:
     include:
       - terraform/*
+
 pool:
   vmImage: ubuntu-latest
 
@@ -22,6 +27,8 @@ stages:
             value: $[variables['System.PullRequest.SourceBranch']]
           - name: INDIVIDUAL_SOURCE
             value: $[variables['Build.SourceBranchName']]
+          - name: IS_TAG
+            value: $[startsWith(variables['Build.SourceBranch'], 'refs/tags/')]
           - name: TARGET
             value: $[variables['System.PullRequest.TargetBranch']]
         steps:
@@ -30,6 +37,9 @@ stages:
           - bash: |
               WORKSPACE=$(python terraform/pipeline/workspace.py)
               echo "##vso[task.setvariable variable=workspace;isOutput=true]$WORKSPACE"
+
+              TAG_TYPE=$(python terraform/pipeline/tag.py)
+              echo "##vso[task.setvariable variable=tag_type;isOutput=true]$TAG_TYPE"
             name: setvars
             displayName: Determine deployment environment
             env:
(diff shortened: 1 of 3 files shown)


angela-tran:
Need to look at this a little closer - specifically the `condition` for the `terraform plan` job step

terraform/pipeline/azure-pipelines.yml
```diff
@@ -70,21 +88,92 @@ stages:
               command: plan
               # wait for lock to be released, in case being used by another pipeline run
               # https://discuss.hashicorp.com/t/terraform-plan-wait-for-lock-to-be-released/6870/2
-              commandOptions: -input=false -lock-timeout=5m
-              workingDirectory: "$(System.DefaultWorkingDirectory)/terraform"
-              # service connection
-              environmentServiceNameAzureRM: deployer
-            # the plan is done as part of the apply (below), so don't bother doing it twice
-            condition: notIn(variables['Build.SourceBranchName'], 'dev', 'test', 'prod')
```
angela-tran, line 78: We no longer need a condition on the `terraform plan` step because it should always run.

angela-tran:
This is ready for review

terraform/pipeline/azure-pipelines.yml
```diff
@@ -70,21 +88,92 @@ stages:
...
+          # need to publish the tfplan to used by next stage if it's going to run
+          - publish: "$(Build.ArtifactStagingDirectory)"
+            displayName: "Publish tfplan for use in TerraformApply"
+            artifact: savedPlan
+            condition: |
+                or(
+                  in(variables['Build.SourceBranchName'], 'dev', 'test', 'prod'),
+                  or(
+                    eq(variables['setvars.tag_type'], 'test'),
+                    eq(variables['setvars.tag_type'], 'prod')
```
thekaveman, line 104: This is great, really clear and easy to see what will need to change when we move away from the branches.

Would it have the same effect to list all the conditionals in a single `or()`, to unnest this double-`or()`?

terraform/pipeline/tag.py
```diff
@@ -0,0 +1,19 @@
+import os
+import re
+
+REASON = os.environ["REASON"]
+# use variable corresponding to tag triggers
+SOURCE = os.environ["INDIVIDUAL_SOURCE"]
+IS_TAG = os.environ["IS_TAG"].lower() == "true"
```
thekaveman, line 7: We're sure this environment variables will always exist with a value right? No need to guard against missing keys?

thekaveman (dismissed):
This looks great @angela-tran!! Huge props for making so much progress in a short amount of time 👍

terraform/pipeline/azure-pipelines.yml
angela-tran: > Would it have the same effect to list all the conditionals in a single `or()`, to unnest this double-`or()`?

Oh lol, yes I think so. For some reason I thought `or()` only allowed 2 arguments, but it actually can take N number ☺️ 

_From [Azure Pipelines - Expressions](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#or)_

> ### or
> 
>    - Evaluates True if any parameter is True
>    - Min parameters: 2. Max parameters: N
>    - Casts parameters to Boolean for evaluation
>    - Short-circuits after first True
>    - Example: or(eq(1, 1), eq(2, 3)) (returns True, short-circuits)

terraform/pipeline/tag.py
angela-tran: Yes, it will always exist because we defined the variable [earlier](https://github.com/cal-itp/eligibility-server/pull/310/files#diff-feb0980fe8b609ae977665357cf3a0d90f0e91fefa84f63ffa70ae660596e5c4R36), and it'll be [automatically inserted into the process environment](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#access-variables-through-the-environment) as an environment variable.

terraform/pipeline/azure-pipelines.yml
angela-tran: Simplified it here: https://github.com/cal-itp/eligibility-server/pull/310/commits/eca6eef93281a535c5559d253667982ddcff13ca

thekaveman (approved):

PR #328, opened 2023-09-26
Fix: match default timeout to benefits

thekaveman:
Makes the default setting more useful and helps avoid the problem we were seeing related to the fix in https://github.com/cal-itp/benefits/pull/1738


angela-tran (approved):
🙌

PR #329, opened 2023-09-27
Feat: deploy for pre-release or release

### LINKED ISSUE #286: Add support for running "Deploy" GitHub workflow when GitHub pre-release/release is created
- [x] Update `deploy.yml` to check for pre-release/release events
- [x] Add step(s) to `deploy.yml` to run appropriate checks (run tests, etc.) if pre-release/release
- [x] Add logic to name the Docker tag appropriately when we see a GitHub pre-release/release
- [x] Create Docker tag with Git tag name too
- [x] Configure [tag protections](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/configuring-tag-protection-rules) so that only benefits-admin team can push tags that trigger deployment (same sort of protection as what we have with branch protections for current deployment model)
- [x] Configure [Required reviewers](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment#required-reviewers) on GitHub deployment environments for test/prod to provide an additional opportunity to check that the deployment looks correct

angela-tran:
Part of #286 

This PR adds to our `Deploy` GitHub workflow so that it can be triggered by a GitHub release (either pre- or full release will work).

These are all the code changes needed to implement #286; the last two tasks on the issue are changes to settings for the GitHub repo.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -7,6 +7,9 @@ on:
       - dev
       - test
       - prod
+  release:
+    types:
+      - published
 
 defaults:
   run:
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -15,9 +15,12 @@ defaults:
   run:
     shell: bash
 
-concurrency:
+env:
   # this expression works like a ternary operation (see https://github.com/actions/runner/issues/409#issuecomment-727565588)
-  group: ${{ github.event_name != 'release' && github.ref_name || github.event.release.prerelease && 'test' || 'prod' }}
+  - DEPLOYMENT_ENVIRONMENT: ${{ github.event_name != 'release' && github.ref_name || github.event.release.prerelease && 'test' || 'prod' }}
+
+concurrency:
+  group: ${{ env.DEPLOYMENT_ENVIRONMENT }}
   cancel-in-progress: true
 
 jobs:
@@ -29,7 +32,7 @@ jobs:
     needs: test
     if: always() && !cancelled() && (github.event_name != 'release' || needs.test.result == 'success')
     runs-on: ubuntu-latest
-    environment: ${{ github.event_name != 'release' && github.ref_name || github.event.release.prerelease && 'test' || 'prod' }}
+    environment: ${{ env.DEPLOYMENT_ENVIRONMENT }}
 
     steps:
       - name: Checkout
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -15,12 +15,15 @@ defaults:
   run:
     shell: bash
 
+concurrency:
+  # this expression works like a ternary operation (see https://github.com/actions/runner/issues/409#issuecomment-727565588)
+  group: ${{ github.event_name != 'release' && github.ref_name || github.event.release.prerelease && 'test' || 'prod' }}
+  cancel-in-progress: true
+
 jobs:
   deploy:
     runs-on: ubuntu-latest
-    # this expression works like a ternary operation (see https://github.com/actions/runner/issues/409#issuecomment-727565588)
     environment: ${{ github.event_name != 'release' && github.ref_name || github.event.release.prerelease && 'test' || 'prod' }}
-    concurrency: ${{ github.event_name != 'release' && github.ref_name || github.event.release.prerelease && 'test' || 'prod' }}
 
     steps:
       - name: Checkout
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -60,5 +60,6 @@ jobs:
           file: Dockerfile
           push: true
           tags: |
+            ghcr.io/${{ github.repository }}:${{ env.DEPLOYMENT_ENVIRONMENT }}
             ghcr.io/${{ github.repository }}:${{ github.ref_name }}
             ghcr.io/${{ github.repository }}:${{ github.sha }}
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-publish.yml
```diff
@@ -44,5 +60,6 @@ jobs:
           file: Dockerfile
           push: true
           tags: |
+            ghcr.io/${{ github.repository }}:${{ env.DEPLOYMENT_ENVIRONMENT }}
             ghcr.io/${{ github.repository }}:${{ github.ref_name }}
             ghcr.io/${{ github.repository }}:${{ github.sha }}
```
angela-tran, line 65: Because we are in an interim state of still doing merge-based deployments, the first two values in this list will be the same... I'm not sure if the `docker/build-push-action@v5` action will allow that.

.github/workflows/docker-publish.yml
```diff
@@ -7,16 +7,32 @@ on:
...
+  cancel-in-progress: true
+
 jobs:
+  test:
+    uses: ./.github/workflows/run-tests.yml
+    if: github.event_name == 'release'
+
   deploy:
+    needs: test
+    if: always() && !cancelled() && (github.event_name != 'release' || needs.test.result == 'success')
```
angela-tran, line 33: `always()` is needed here because even if `test` didn't run, we still want this to run. 
`!cancelled()` is needed because if the whole workflow was cancelled, we don't want this to run.
`github.event_name != 'release' || needs.test.result == 'success')` is needed because if `test` _did_ run, we only want this to run if `test` succeeded.

.github/workflows/docker-publish.yml
angela-tran: There's not a really easy way to test this without just merging this and trying a merge-based deployment.

If we really do want to test this before merging though, it could be set up in a smaller experimental repo. It just takes some time to get it all set up.

angela-tran:
This is ready for review, but note that this is a stretch item for the sprint so it's ok if this doesn't get reviewed for a while

.github/workflows/docker-publish.yml
thekaveman: I'm not following the need for `always()` -- isn't that like saying `true AND x` which evaluates to whatever `x` is?

angela-tran: I think you are right that it is redundant! I tried it out myself to verify.

Removed it in 659c5c80f9fc658718c8fc1abb55b2c8c42c83f2

.github/workflows/docker-publish.yml
```diff
@@ -7,16 +7,32 @@ on:
...
+  cancel-in-progress: true
+
 jobs:
+  test:
+    uses: ./.github/workflows/run-tests.yml
+    if: github.event_name == 'release'
+
   deploy:
+    needs: test
+    if: (!cancelled() && (github.event_name != 'release' || needs.test.result == 'success'))
```
thekaveman, line 33: Great!

I think it would be useful to have a little comment here describing what you [said in this PR comment](https://github.com/cal-itp/eligibility-server/pull/329#discussion_r1341766075)

thekaveman (changes requested):

.github/workflows/docker-publish.yml
```diff
@@ -7,16 +7,32 @@ on:
...
+    types:
+      - published
 
 defaults:
   run:
     shell: bash
 
+env:
+  # this expression works like a ternary operation (see https://github.com/actions/runner/issues/409#issuecomment-727565588)
+  - DEPLOYMENT_ENVIRONMENT: ${{ github.event_name != 'release' && github.ref_name || github.event.release.prerelease && 'test' || 'prod' }}
```
thekaveman, line 20: Finally understand this 🤯 😨

thekaveman (approved):
LGTM!

I just ran this in my terminal, which makes me think the duplicate tag here won't be a problem:

```console
$ docker build --no-cache -t eligibility_server:main -t eligibility_server:main .
[+] Building 0.0s (0/0)                                                                                                                                              docker:default
[+] Building 46.0s (11/11) FINISHED                                                                                                                                  docker:default 
 => [internal] load build definition from Dockerfile                                                                                                                           6.0s 
 => => transferring dockerfile: 404B                                                                                                                                           0.0s 
 => [internal] load .dockerignore                                                                                                                                              6.4s 
 => => transferring context: 81B                                                                                                                                               0.0s 
 => [internal] load metadata for ghcr.io/cal-itp/docker-python-web:main                                                                                                        1.9s 
 => CACHED [1/6] FROM ghcr.io/cal-itp/docker-python-web:main@sha256:7a8a6bc350c5ebffd04b3855912abeb3fb8a1b23c8578ad47c4a463a16a6070c                                           0.0s 
 => [internal] load build context                                                                                                                                              0.8s
 => => transferring context: 674B                                                                                                                                              0.0s
 => [2/6] RUN python -m pip install --upgrade pip                                                                                                                              7.0s
 => [3/6] COPY bin bin                                                                                                                                                         0.4s
 => [4/6] COPY eligibility_server/ eligibility_server/                                                                                                                         0.3s
 => [5/6] COPY pyproject.toml pyproject.toml                                                                                                                                   0.2s
 => [6/6] RUN pip install -e .                                                                                                                                                28.1s
 => exporting to image                                                                                                                                                         0.8s 
 => => exporting layers                                                                                                                                                        0.7s 
 => => writing image sha256:97398d05669bda9f05d9a655dbd1f1e0b6c95aa5c22ba61e5e45a0d11df5c09e                                                                                   0.0s 
 => => naming to docker.io/library/eligibility_server:main
```

angela-tran:
> I just ran this in my terminal, which makes me think the duplicate tag here won't be a problem

Ooh, good idea to try that 👍

Issue #334, opened 2023-09-27
Make git SHA and pipdeptree available as static files


thekaveman:
Similar to e.g. `benefits.calitp.org/static/sha.txt`, it would be nice to be able to see the SHA of the deployed version of Eligibility Server as another quick check.

See also the work in cal-itp/benefits#1733 to get the Python dependency tree available as a static file in the image.

* [ ] Need to figure out how to serve static files from Eligibility Server -- ideally we aren't creating endpoints in the app, and this can be handled by `nginx` before the request even gets to the app (like how static files work in `benefits`).

angela-tran:
From the [Flask docs on static files](https://flask.palletsprojects.com/en/2.3.x/tutorial/static/), it is as simple as having a file in the `/static` directory.

I tested it out by creating a file at `eligibility_server/static/sha.txt` and launching the Flask app. I was able to see the file at `<url>/static/sha.txt` 🙂

PR #335, opened 2023-09-28
Feat: make git hash available in static file

angela-tran:
Part of #334


thekaveman:
It looks like #336 has both static files? Should I review this one or should we close this one?

angela-tran:
@thekaveman I think review this one and then I'll rebase the other one

thekaveman (approved):

PR #336, opened 2023-09-28
Feat: pipdeptree

### LINKED ISSUE #334: Make git SHA and pipdeptree available as static files

angela-tran:
Closes #334 

Similar to cal-itp/benefits#1733

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -22,6 +22,21 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
+      - uses: actions/setup-python@v4
+        with:
+          python-version-file: .github/workflows/.python-version
+          cache: pip
+          cache-dependency-path: "**/pyproject.toml"
+
+      - name: Write python packages to file
+        run: |
+          python -m venv .venv
+          source .venv/bin/activate
+          pip install pipdeptree
+          pip install -e .
+          pipdeptree
+          pipdeptree >> eligibility_server/static/requirements.txt
+
       - name: Write commit SHA to file
         run: echo "${{ github.sha }}" >> eligibility_server/static/sha.txt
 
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-publish.yml
```diff
@@ -22,6 +22,21 @@ jobs:
...
+          cache-dependency-path: "**/pyproject.toml"
+
+      - name: Write python packages to file
+        run: |
+          python -m venv .venv
+          source .venv/bin/activate
+          pip install pipdeptree
+          pip install -e .
+          pipdeptree
+          pipdeptree >> benefits/static/requirements.txt
```
angela-tran, line 38: ah need to change this

thekaveman (dismissed):
Approving this, not sure why the other one failed though? 

https://github.com/cal-itp/eligibility-server/actions/runs/6342792432/job/17229278668

https://www.gnu.org/software/bash/manual/html_node/Redirections.html#Appending-Redirected-Output

> ... If the file does not exist it is created.

angela-tran:
@thekaveman Hmm, maybe it'll create the file if it doesn't exist but not necessarily any parent directories. I'll test out this workflow more - gonna try running it locally with `act`!

angela-tran:
I was able to reproduce the issue with `act` 😍 

image

I think this means we need to add a step to create the `static` directory so that it can have both `sha.txt` and `requirements.txt` in it.

For benefits we don't need this step since we already have a `static` directory with other stuff in it

angela-tran:
With the step added in c3c010c, I got `act` to successfully write to `eligibility_server/static/sha.txt` and `eligibility_server/static/requirements.txt`, so hopefully it works in GitHub too!

image

thekaveman (approved):
Nice work! 🙏 `act`

PR #339, opened 2023-10-02
Fix: inline expression for deployment environment

angela-tran:
See the [failed workflow run](https://github.com/cal-itp/eligibility-server/actions/runs/6382604530) from when I merged #329 into `dev`.

It turns out that the `env` context is not available in the places I wanted to use it (`concurrency`, `jobs.<job>.environment`), so in this PR, I put the expression back as inline instead of having it be a variable.

See [GitHub Actions docs](https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability) for more on where different contexts are available.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -15,12 +15,9 @@ defaults:
   run:
     shell: bash
 
-env:
-  # this expression works like a ternary operation (see https://github.com/actions/runner/issues/409#issuecomment-727565588)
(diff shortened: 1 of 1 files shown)


angela-tran:
Another problem that came out of trying to define and use `env.DEPLOYMENT_ENVIRONMENT` is that `act` for some reason was not able to get past the `!` character that was in the variable's expression. The error I was getting seems to be a Go error, and I think is a bug in `act`. GitHub was fine with defining the variable.

image

(see https://github.com/nektos/act/issues/284 for a past bug that seems similar(?))

Because of this, I wasn't able to make use of `act` to validate my workflow.

Now that I've in-lined the expression, I do get a successful dry run, so that seems promising.

image

thekaveman (approved):
Bummer!

PR #345, opened 2023-10-10
Chore: Terraform code cleanup

angela-tran:
Follow-up to #343 

Fix formatting and comment


angela-tran:
Also using this PR to test an idea for passing variables to Terraform config. 

The idea:
 - pass filepath of variable template as a runtime parameter
 - use variable template to load variables as environment variables
 - Terraform automatically sees those variables because of their `TF_VAR` prefix

thekaveman (approved):

Issue #296, opened 2023-08-16
SBMTD Azure setup


thekaveman:
Some tasks may need help from SBMTD IT.

Most of these tasks come from the [notes doc when we did this with MST](https://docs.google.com/document/d/12uzuKyvyabHAOaeQc6k2jQIG5pQprdEyBpfST_dY2ME/edit).

## Azure basics

- [x] Create Azure user group for Compiler, roles:
  - `Application developer`
- [x] Add @angela-tran, @machikoyasuda, @thekaveman to group
- [x] Create Subscription, assign Compiler group the `Contributor` role

## Azure DevOps

- [x] Create Azure DevOps organization
- [x] Install Terraform extension on organization
- [x] Create Azure DevOps project
- [x] Turn off services not in use (`Project settings > Overview > services`) -- only Pipelines is used
- [x] Add Compiler users (ideally via group)
- [x] Create service connection
  - Type: Azure Resource Manager
  - Method: Service principal (automatic)
  - https://learn.microsoft.com/en-us/training/modules/authenticate-azure-deployment-pipeline-service-principals/
- [x] Create pipeline by importing `azure-pipelines.yml` from this repo
  - Depends on #291 
- [x] Set environment variables
  - [x] [`TF_IN_AUTOMATION=1`](https://developer.hashicorp.com/terraform/tutorials/automation/automate-terraform#controlling-terraform-output-in-automation)
  - [x] `TF_VAR_DEPLOYER_APP_OBJECT_ID`
  - [x] `TF_VAR_ENGINEERING_GROUP_OBJECT_ID`
  - [x] `TF_VAR_AGENCY_CARD_DATA_ETL_APP_OBJECT_ID` (renamed from `TF_VAR_VELOCITY_ETL_APP_OBJECT_ID`)
  - [x] `TF_VAR_IP_ADDRESS_WHITELIST_DEV/TEST/PROD`
- [x] Create Environment named "Approval"
- [x] #362

## Terraform

- [x] (Manual) Create Resource Group and storage account dedicated to the Terraform state: `sbmtd-mobility-pass-tf`
- [x] (Manual) Create container in storage account named `tfstate`
- [x] (Manual x3 environments) Create environment Resource Group, Region: West US
  - [x] `sbmtd-mobility-pass-eligibility-dev`
  - [x] `sbmtd-mobility-pass-eligibility-test`
  - [x] `sbmtd-mobility-pass-eligibility-prod`
[...]

PR #349, opened 2023-10-12
Feat: SBMTD infra pipeline

angela-tran:
Part of #296

### DIFF
--- a/terraform/sbmtd/azure-vars.yml
+++ b/terraform/sbmtd/azure-vars.yml
@@ -1,0 +1,7 @@
+variables:
+  TF_RESOURCE_GROUP: "sbmtd-mobility-pass-tf"
+  TF_STORAGE_ACCOUNT: "sbmtdmobilitypasstf"
+  TF_VAR_AGENCY_CARD: "sbmtd-mobility-pass"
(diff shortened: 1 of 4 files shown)


thekaveman (approved):

terraform/sbmtd/azure-vars.yml
```diff
@@ -0,0 +1,7 @@
+variables:
+  TF_RESOURCE_GROUP: "sbmtd-mobility-pass-tf"
+  TF_STORAGE_ACCOUNT: "sbmtdmobilitypasstf"
+  TF_VAR_AGENCY_CARD: "sbmtd-mobility-pass"
+  TF_VAR_AGENCY_RESOURCE_GROUP_PREFIX: "sbmtd-mobility-pass"
+  TF_VAR_AGENCY_STORAGE_ACCOUNT_PREFIX: "sbmtdmobilitypass"
+  TF_VAR_AGENCY_CARD_DATA_ETL_FILE: "mobilitypass.csv"
```
angela-tran, line 7: We need this to match whatever the ETL scripts want this to be named

thekaveman: I think we can make the ETL script match this, or whatever else we need, since that is not technically created or running yet.

PR #354, opened 2023-10-25
Fix: use agency card as prefix for key vault name

### LINKED ISSUE #353: Key Vault name needs to be globally unique
Follow-up to #349 

From https://dev.azure.com/sbmtd/eligibility-server/_build/results?buildId=12&view=logs&j=de9ad78b-21c3-52d3-f75b-ac0998b9a9c9&t=9f0d2323-8cb9-5ca6-9fa8-c6837cb4c41d

> ╷
> │ `Error: creating Vault: (Name "eligibility-server-dev" / Resource Group "sbmtd-mobility-pass-eligibility-dev"): keyvault.VaultsClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: autorest/azure: Service returned an error. Status=<nil> Code="VaultAlreadyExists" Message="The vault name 'eligibility-server-dev' is already in use. Vault names are globally unique so it is possible that the name is already taken. If you are sure that the vault name was not taken then it is possible that a vault with the same name was recently deleted but not purged after being placed in a recoverable state. If the vault is in a recoverable state then the vault will need to be purged before reusing the name. For more information about VaultAlreadyExists, soft delete and purging a vault follow this link [https://go.microsoft.com/fwlink/?linkid=2147740."](https://go.microsoft.com/fwlink/?linkid=2147740.%22)`
> │ 
> │   with azurerm_key_vault.main,
> │   on key_vault.tf line 1, in resource "azurerm_key_vault" "main":
> │    1: resource "azurerm_key_vault" "main" {
> │ 
> ╵

Here are the Resource naming restrictions for Azure Key Vaults: https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftkeyvault

> Length: 3-24
> 
> Valid characters:
> Alphanumerics and hyphens.
> Start with letter. End with letter or digit. Can't contain consecutive hyphens.

angela-tran:
Closes #353 

Use the agency card name instead of a hard-coded `eligibility-server`. Key Vault names are limited to 24 characters.

Note that this will rename MST's key vault, which is ok because it just contains the Slack email for notifications.

### Post-approval
- [x] Manually delete MST key vault since `lifecycle.prevent_destroy` is set to `true`

### Post-merge
- [ ] Add back slack email secret for MST
- [ ] Add slack email secret for SBMTD

### DIFF
--- a/terraform/key_vault.tf
+++ b/terraform/key_vault.tf
@@ -1,6 +1,6 @@
 resource "azurerm_key_vault" "main" {
   # name needs to be globally unique
-  name                = "eligibility-server-${local.env_name}"
+  name                = "${var.AGENCY_CARD}-${local.env_name}"
   location            = data.azurerm_resource_group.main.location
   resource_group_name = data.azurerm_resource_group.main.name
   sku_name            = "standard"


thekaveman:
So it seems like the `plan` failed because of 

> Manually delete MST key vault since `lifecycle.prevent_destroy` is set to `true`

angela-tran:
> So it seems like the `plan` failed because of
> 
> > Manually delete MST key vault since `lifecycle.prevent_destroy` is set to `true`

Good point that you probably want to see a successful plan before approving. I can go ahead and delete the Key Vault in dev.

thekaveman:
Ah, I see -- you were expecting this and would clean it up post merge. It's all good :+1:

thekaveman (approved):

angela-tran:
Hmm, we've made it very hard to fully delete our Key Vaults. We enabled purge protection on them, and once it's enabled, you can't disable it. This means when we delete a Key Vault, it is in a soft-deleted state and can be recovered within the configured retention period (ours is set to 90 days). You cannot purge the Key Vault until the retention period has passed.

Here's a screenshot of my attempt at deleting + purging the MST Key Vault in dev. I recovered the Key Vault when I realized I can't actually purge it.

image

thekaveman:
> You cannot purge the Key Vault until the retention period has passed.

Is it possible to shorten the retention period for Eligibility Server key vaults? Like to 1 day? Via Terraform or otherwise

angela-tran:
> Is it possible to shorten the retention period for Eligibility Server key vaults? Like to 1 day? Via Terraform or otherwise

No, we can't change the retention period once it's set. 

I just realized though that it's ok for the Key Vault to be in a soft-deleted state... we're not trying to reuse that name; we just want it to not exist so that Terraform doesn't try to delete it. 🧠 I'm gonna try again with deleting it and re-running the pipeline to get a Terraform plan

PR #253, opened 2023-04-13
Feat: Sentry error monitoring

angela-tran:
Closes #252 
Closes #312 

## How to test / review

Create a new Python settings file in your local `config/` directory, e.g. `config/sentry.py`, with the following contents:

```python
# get the DSN from https://sentry.calitp.org/settings/sentry/projects/eligibility-server/keys/
SENTRY_DSN = "https://..."  

# make up a name for testing
AGENCY_NAME = "Your Transit Agency"
```

In your local `.env` file, make sure to point to your custom settings file: 

```env
ELIGIBILITY_SERVER_SETTINGS=../config/sentry.py
```

Get this branch and rebuild and reopen in the devcontainer.

`F5` to run the app

Go to `http://localhost:PORT/error` to trigger a Sentry error

Log in to Sentry to confirm your details were recorded in the error.

## TODOs
 - [x] ~Figure out / decide on a way to [single-source the version number](https://packaging.python.org/en/latest/guides/single-sourcing-package-version/) and read it when setting `release` for Sentry~ determined that we don't need this. Benefits uses `VERSION` as a fallback, but in practice, we're always getting the git hash.

## Before merging
 - [x] Set `SENTRY_DSN` as a setting for current deployments of `eligibility-server` (it will be within the file that is specified by [`ELIGIBILITY_SERVER_SETTINGS`](https://github.com/cal-itp/eligibility-server/blob/3052d6f43ff3b75f7ea0b2011f95f073885a872b/terraform/app_service.tf#L52))
    - [x] dev
    - [x] test
    - [x] prod
 - [x] Set `AGENCY_NAME` as a setting for current deployments of `eligibility-server` (it will be within the file that is specified by [`ELIGIBILITY_SERVER_SETTINGS`](https://github.com/cal-itp/eligibility-server/blob/3052d6f43ff3b75f7ea0b2011f95f073885a872b/terraform/app_service.tf#L52)). This should match the corresponding [`agency.long_name` from Benefits](https://github.com/cal-itp/benefits/blob/dev/benefits/core/migrations/0002_data.py#L286)
    - [x] dev
    - [x] test
    - [x] prod

PR #355, opened 2023-10-26
Deploy to test

angela-tran:


thekaveman:
Can we make #253 go as part of this cycle?

thekaveman:
@angela-tran the `plan` failed, is that expected?

> Error: Instance cannot be destroyed
│ 
│   on key_vault.tf line 1:
│    1: resource "azurerm_key_vault" "main" {
│ 
│ Resource azurerm_key_vault.main has lifecycle.prevent_destroy set, but the
│ plan calls for this resource to be destroyed.

Seems like the same message as before?

thekaveman:
Is this where we need to soft-delete the key vault and not worry since we're not trying to reuse the name anyway?

angela-tran:
> @angela-tran the `plan` failed, is that expected?
> 
> > Error: Instance cannot be destroyed
> > │
> > │   on key_vault.tf line 1:
> > │    1: resource "azurerm_key_vault" "main" {
> > │
> > │ Resource azurerm_key_vault.main has lifecycle.prevent_destroy set, but the
> > │ plan calls for this resource to be destroyed.
> 
> Seems like the same message as before?

Right, it's the same message as when we `plan`'d the change in #354 for `dev`. So now we need to do the same process of manually deleting the Key Vault (this time the one named `eligibility-server-test`) and setting the secret again in the new `mst-courtesy-card-test` Key Vault

thekaveman (approved):

Issue #297, opened 2023-08-16
SBMTD Agency card server setup


thekaveman:
Depends on #296 

Based on [notes from when we did this for MST](https://docs.google.com/document/d/12uzuKyvyabHAOaeQc6k2jQIG5pQprdEyBpfST_dY2ME/edit).

## Azure

- [x] For `prod`, create service principal for upload job
  - https://learn.microsoft.com/en-us/azure/active-directory/develop/howto-create-service-principal-portal#app-registration-app-objects-and-service-principals
  - https://learn.microsoft.com/en-us/cli/azure/authenticate-azure-cli#sign-in-with-a-service-principal
  - https://learn.microsoft.com/en-us/rest/api/storageservices/authorize-with-azure-active-directory#permissions-for-calling-blob-and-queue-data-operations
- [x] Create client secret for principal
  - Username is the Application (client) ID
  - Password is the secret
  - Take note of expiration date on the client secret
 - [x] Give role assignment in Azure portal to ETL service principal in `prod` as documented in infrastructure README

## Agency card server

- [x] Install [Azure CLI](https://learn.microsoft.com/en-us/cli/azure/install-azure-cli)
- [x] Confirm login command:
 ```bash
 az login --service-principal -u <object id> -p <secret> --tenant <tenant id>
 ```
- [ ] Confirm upload command: 
```bash
az storage blob upload -f <file.txt> --account-name <account> \
  -c <storage container> --name <file.txt> --auth-mode login --overwrite
```
- [x] Write the above commands to a file named `load.bat`
- [x] Determine table structure and query for Agency card database, e.g. 
```sql
SELECT U.MobilityPassId as sub, U.LastName as name, 'mobility_pass' AS type
FROM Users U INNER JOIN UserCredentials UC ON U.MobilityPassId = UC.MobilityPassId 
WHERE UC.Active = 1 AND U.LastName <> ''
ORDER BY U.LastName, U.MobilityPassId
```
- [x] Use `bcp` to extract records into UTF-8 encoded (codepoint 65001) CSV with a header row in a file `extract.bat`:
```cmd
bcp "SELECT ..." queryout data.txt -T -c -C 65001 -d <DATABASE> -S <SERVER> -t ,

echo sub,name,type >> temp.txt
type data.txt >> temp.txt
del data.txt
ren temp.txt data.txt
```
- [x] Install [hashfields](https://github.com/cal-itp/hashfields/releases) ~(and potentially Python as well)~
[...]

PR #359, opened 2023-10-30
Refactor: remove role assignment from Terraform

angela-tran:
Part of #296 

We found that because the pipeline service principal defaults to having `Contributor` role, it is not able to create the role assignment for the ETL principal on the Storage Container. 

After some discussion, we decided we should remove the role assignment from being managed by Terraform / the pipeline, and it'll be a part of the manual creation steps to set up the role assignment.

This PR deletes the role configuration from Terraform and adds documentation for the manual steps.

### Pre-merge / post-merge actions
For **MST**, we need to 
 - [x] run `terraform state rm azurerm_role_assignment.velocity_etl[0]` so that Terraform "forgets" about the role and doesn't try to destroy it (see [docs on `rm`](https://developer.hashicorp.com/terraform/cli/commands/state/rm)). 
 - After merging, the `prod` pipeline should succeed, and there's nothing left to do.

For **SBMTD**, 
 - [x] there's no pre-merge action to do. 
 - After merging, the `prod` pipeline should succeed since now there is no permissions error to run into. The last thing to do is to manually create the role assignment for the ETL service principal. Doing so requires the `Owner` role on your Azure account in the SBMTD subscription. (This has been added as a task on #297)


angela-tran:
There's a little more code cleanup to do

terraform/variables.tf
```diff
@@ -8,11 +8,6 @@ variable "AGENCY_CARD" {
   type        = string
 }
 
-variable "AGENCY_CARD_DATA_ETL_APP_OBJECT_ID" {
-  description = "Object ID from the registered application for the Agency Card server ETL uploading: https://cloudsight.zendesk.com/hc/en-us/articles/360016785598-Azure-finding-your-service-principal-object-ID"
-  type        = string
-}
-
 variable "AGENCY_CARD_DATA_ETL_FILE" {
```
thekaveman, line 11: I don't think we need this variable anymore either right?

thekaveman (changes requested):

thekaveman (changes requested):
And let's also delete the unused variables from the agency-specific `azure-vars.yml` files.

terraform/variables.tf
angela-tran: Thanks for catching! I removed the variable and values in https://github.com/cal-itp/eligibility-server/pull/359/commits/83d5231d7f9510986f83ce3ed1de939c903c4683

angela-tran:
Successfully ran the `terraform state rm...` command for MST prod workspace:
```
C:\dev\eligibility-server\terraform>terraform state rm azurerm_role_assignment.velocity_etl[0]
Removed azurerm_role_assignment.velocity_etl[0]
Successfully removed 1 resource instance(s).
```

thekaveman (approved):

Issue #294, opened 2023-08-16
Trigger azure-pipelines.yml via GitHub Action


thekaveman:
This has some overlap with #284 and so will need to coordinate, possibly building on top of the work done there.

A single GitHub repository can be associated with multiple Azure DevOps organizations/projects:

> Once the GitHub App is installed, pipelines can be created for the organization's repositories in different Azure DevOps organizations and projects.

_From [Azure docs: Build GitHub repositories](https://learn.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#create-pipelines-in-multiple-azure-devops-organizations-and-projects)_.

But the docs go on to say (emphasis mine):

> However, if you create pipelines for a single repository in multiple Azure DevOps organizations, **_only the first organization's pipelines can be automatically triggered by GitHub commits or pull requests_**. Manual or scheduled builds are still possible in secondary Azure DevOps organizations.

So in order to deploy to multiple Azure DevOps orgs (i.e. MST and SBMTD), we're going to need to trigger the `azure-pipelines.yml` run manually, via GitHub Actions. 

Further, (if possible) it would make sense to trigger every org's pipeline in the same way, e.g. _not_ have one of the DevOps orgs setup for "automatic trigger by GitHub commits or pull requests" like mentioned in the docs above. Each org should deploy via the GitHub Action.

Note _manually_ here does not mean literally a human clicking a button to trigger a workflow. It means we'll need a GitHub Actions workflow (that acts on e.g. tag, to accommodate what is needed for #284) that then triggers the pipeline.

angela-tran:
This task may not be necessary based on some findings from my experiments documented in https://github.com/cal-itp/eligibility-server/issues/284#issuecomment-1688736156

angela-tran:
Once #346 is merged in, I suggest that as our next step we set up an Azure DevOps pipeline in SBMTD's organization and see if GitHub commits / pull requests automatically trigger in both MST and SBMTD or just one. That will help us decide if we need to do this ticket or not.

angela-tran:
I ran several experiments to help us decide if we need to do this ticket or not. In my first experiment, I made a test commit - 750d85f - on the `feat/sbmtd-infra-pipeline` branch. In my second experiment, I opened #352. Both experiments resulted in only triggering the MST pipeline.

This behavior matches what the Azure docs say. 

[...]

PR #361, opened 2023-10-31
Deploy to prod

angela-tran:
This is to get #359 into `prod`


thekaveman (approved):
I think we're good to merge this.

But it seems like we will need to look into #294 to at least get the `plan` running in each pipeline when PRs are opened.

PR #378, opened 2023-12-05
Refactor: alert suppression rule block

### LINKED ISSUE #356: Express uptime supression window in Pacific time
This was originally expressed in [Eastern Standard Time](https://github.com/cal-itp/eligibility-server/blob/dev/terraform/suppress.arm.json#L24) for... reasons... 🤷 

But it introduces confusion and complexity, since the whole point of the suppression is for data loading according to a _California transit agency_ e.g. Pacific time zone.

thekaveman:
Redefine directly in a Terraform block, which is more straightforward and clear than the ARM template.

Reference: [`azurerm_monitor_alert_processing_rule_suppression`](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/monitor_alert_processing_rule_suppression)

The [`scope`](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/monitor_alert_processing_rule_suppression#scopes) here is set directly to the ID of the healthcheck alert, hence this suppression rule targets only that single resource. Thus we do not need any of the filtering for e.g. `target_resource_type`.

Closes #356


angela-tran (approved):
I checked that these attributes are in [`azurerm 3.28.0`](https://registry.terraform.io/providers/hashicorp/azurerm/3.28.0/docs/resources/monitor_alert_processing_rule_suppression#recurrence) (the version we have in our [lock](https://github.com/cal-itp/eligibility-server/blob/dev/terraform/.terraform.lock.hcl#L5) file). This looks good to me 👍

PR #381, opened 2023-12-06
Breaking change: Remove JSON file support

### LINKED ISSUE #348: Clean up old JSON import code
This was used early on for local testing, but in production we deal with CSV files.

Let's clean up the code so that the CSV path is the only path, and ensure all local/testing files also only use this path.

machikoyasuda:
closes #348 

## What this PR does
- Remove JSON import method
- Remove JSON from Documentation
- Remove JSON test file


thekaveman (approved):
This is great! :broom: :broom:

PR #392, opened 2023-12-19
Fix: build Docker image with unshallow copy

### LINKED ISSUE #391: Ensure the Docker image built by GitHub Actions behaves similar to one built locally
After troubleshooting and fixing the issue causing our Azure app service to not start (see #389 and #390), I noticed the version in the app service's Docker container is not correct. I then realized a huge potential reason why I haven't been able to reproduce any of the problems: I was testing with a Docker image that I locally built, and Azure is using one built by our GitHub Action!

I downloaded [the image from GHCR](https://github.com/cal-itp/eligibility-server/pkgs/container/eligibility-server/159796098?tag=96e8f16aebeb70c1a917bf2c7e86afcbd05686b7) and see the same incorrect version.

Looking at the workflow logs, I see there is a [warning about a shallow copy](https://github.com/cal-itp/eligibility-server/actions/runs/7267804433/job/19802386849#step:9:308). This issue seems relevant: https://github.com/pypa/setuptools_scm/issues/414

angela-tran:
Closes #391

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -37,6 +37,11 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
+      - name: Get history and tags for SCM versioning to work
+        run: |
+          git fetch --prune --unshallow
+          git fetch --depth=1 origin +refs/tags/*:refs/tags/*
+
       - uses: actions/setup-python@v5
         with:
           python-version-file: .github/workflows/.python-version


machikoyasuda (approved):

angela-tran:
This worked!

From Azure for dev instance:
```
023-12-20T22:22:27.401681466Z + bin/init.sh
2023-12-20T22:22:27.462608121Z + flask init-db
2023-12-20T22:22:35.993407299Z SENTRY_DSN not set, so won't send events
2023-12-20T22:22:35.993438699Z [2023-12-20 22:22:35,993] INFO eligibility_server.app:31 Starting Eligibility Server 2023.11.1rc2.dev37+g0f1456f
```

thekaveman:
@angela-tran this was a great find and fix!

I think it can be done a bit simpler directly from the [`actions/checkout@v4` step](https://github.com/actions/checkout?tab=readme-ov-file#usage):

```yml
- uses: actions/checkout@v4
  with:
    # Number of commits to fetch. 0 indicates all history for all branches and tags.
    # Default: 1
    fetch-depth: 0 
    
    # Whether to fetch tags, even if fetch-depth > 0.
    # Default: false
    fetch-tags: true
```

Maybe the `fetch-depth` doesn't matter as much when we force `fetch-tags`, which would be even simpler!

PR #394, opened 2023-12-22
Feat: create release for tag

### LINKED ISSUE #393: Automate creation of GitHub pre-release/release when Git tag is pushed with specific name
When creating tasks for #282, we determined it was easier to have the Azure DevOps pipeline and the "Deploy" GitHub workflow be kicked off from a git tag push instead of from a GitHub pre-release/release. (This was mainly because Azure DevOps doesn't provide a way to listen for GitHub events.)

To minimize the number of manual steps that need to be done, we should make it so the push of a Git tag with the right naming pattern will automatically trigger the creation of a GitHub pre-release/release.

angela-tran:
Closes #393


.github/workflows/release.yml
```diff
@@ -0,0 +1,23 @@
+on:
+    push:
+      tags:
+      # pre-release tag
+      - "202[3-9].[0-9][0-9].[0-9]+-rc[0-9]+"
```
angela-tran, line 5: Let's not create a pre-release for these (so remove this trigger)

angela-tran: Simplified the workflow in [3517e26](https://github.com/cal-itp/eligibility-server/pull/394/commits/3517e2625a7d9d44fe6fd8cee82f9bb2079eb4e8)

.github/workflows/docker-publish.yml
```diff
@@ -7,31 +7,33 @@ on:
...
+      - "202[3-9].[0-9][0-9].[0-9]+"
 
 defaults:
   run:
     shell: bash
 
 concurrency:
   # this expression gives us the name of the deployment environment. It works like a ternary operation (see https://github.com/actions/runner/issues/409#issuecomment-727565588)
-  group: ${{ github.event_name != 'release' && github.ref_name || github.event.release.prerelease && 'test' || 'prod' }}
+  group: ${{ !startsWith(github.ref, 'refs/tags/') && github.ref_name || contains(github.ref, '-rc') && 'test' || 'prod' }}
```
thekaveman, line 22: Would it make this workflow a little more clear if these expressions were captured as re-usable [environment variables](https://docs.github.com/en/actions/learn-github-actions/variables)?

.github/workflows/release.yml
```diff
@@ -0,0 +1,21 @@
+on:
+    push:
```
thekaveman, line 2: Looks like some extra space here

thekaveman (approved):
Non-blocking comments. Looks good!

.github/workflows/docker-publish.yml
angela-tran: It would definitely make it clearer, but unfortunately the `env` context is not available under `concurrency` or other sections where we'd want to use the variable.

[#339](https://github.com/cal-itp/eligibility-server/pull/339/files) has some more detail on when we tried capturing these expressions as an environment variable.

.github/workflows/release.yml
angela-tran: Thanks, fixed it in https://github.com/cal-itp/eligibility-server/pull/394/commits/c521775ff6c7342bdf31ab419a565158e88e78e8

thekaveman (approved):

Issue #288, opened 2023-08-08
Do a release with new deployment model


angela-tran:
Depends on #284, #286, #287 

 - **Deploy an `-rc` version to test environment**
   - [x] Check out the default branch and make sure it's up-to-date with remote default branch
   - [x] Make an annotated Git tag following release candidate version number format, `2024.01.1-rc1`, and push it
   - Check on the Azure DevOps infra pipeline for each transit agency:
      - [x] Did the pipeline run?
      - [x] Was `terraform plan` successful?
      - [x] Did the pipeline wait for your approval to run `terraform apply`?
      - [x] Was `terraform apply` succcessful?
   - Check on the GitHub workflow that does the Docker publish:
      - [x] Did the tests run?
      - [x] Did a new Docker image get published with the expected Docker tags (`2024.01.1-rc1`, `test`, and the Git SHA)?
   - QA the app:
      - [x] Does the app respond? (`/healthcheck`)
         - The app failed to start up because of an error in the `eligibility_server.sentry` module - fixed in #398 
      - [x] Is the version number correct?
 - **Deploy a new version to prod environment**
   - [x] Check out the release candidate tag
   - [x] Make an annotated Git tag following release version number format, `2024.01.1`, and push it
      - `git tag -a 2024.01.1`
      - `git push origin 2024.01.1`
   - Check on the Azure DevOps infra pipeline for each transit agency:
      - [x] Did the pipeline run?
      - [x] Was `terraform plan` successful?
      - [x] Did the pipeline wait for your approval to run `terraform apply`?
      - [x] Was `terraform apply` succcessful?
   - Check on the GitHub workflow that does the Docker publish:
      - [x] Did the tests run?
      - [x] Did a new Docker image get published with the expected Docker tags (`2024.01.1`, `prod`, and the Git SHA)?
   - Check on the GitHub workflow that creates the GitHub release:
      - [x] Was a release created for the right tag?
      - [x] Were release notes generated?
         - Found an issue with release notes - it diffs against the very beginning to generate release notes which is incorrect
            - Attempting to fix with #402 and #403
            - Merge to `dev` went through correctly ✅ 
            - Tag push for release candidate `2024.01.3-rc1` 
[...]

PR #405, opened 2024-01-09
Fix: make path filter work for tags

angela-tran:
While working on #288, we noticed that even though our CI triggers in `mst/azure-pipelines.yml` and `sbmtd/azure-pipelines.yml` specify that the pipeline should only run if files under `/terraform` are changed, the infra pipeline ran for `2024.01.3-rc1`, which did not contain any `/terraform` changes.

Reading over the [Azure docs](https://learn.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#paths) more, we realized that `path` filters only combine with the triggers under `branches`.

This PR refactors the tag patterns so that they're specified under the `branches` section.

### DIFF
--- a/terraform/mst/azure-pipelines.yml
+++ b/terraform/mst/azure-pipelines.yml
@@ -4,10 +4,8 @@ trigger:
       - dev
       - test
       - prod
-  tags:
-    include:
-      - 20??.??.?*-rc?*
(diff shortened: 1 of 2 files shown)


thekaveman (approved):
Nice quick fix!

PR #412, opened 2024-01-25
Fix: don't use docker build cache in CI

thekaveman:
Follow up to #411 since we are still seeing a cached package version during `docker-publish`.


thekaveman:
Cowboy time, merging without waiting for approval :cowboy_hat_face: :grimacing:

PR #463, opened 2024-06-11
Refactor: move the application directory up a level

thekaveman:
Related to:

* https://github.com/cal-itp/docker-python-web/pull/42
* https://github.com/cal-itp/benefits/pull/2156

This PR should be merged after the one in docker-python-web.

This PR also removes the explicit `WORKDIR` setting in the appcontainer Dockerfile, as the base image already sets the default we want.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -4,7 +4,7 @@
   "dockerComposeFile": ["../compose.yml"],
   "service": "dev",
   "runServices": ["dev", "docs"],
-  "workspaceFolder": "/home/calitp/app",
+  "workspaceFolder": "/calitp/app",
   "postAttachCommand": ["/bin/bash", ".devcontainer/postAttach.sh"],
   "postStartCommand": ["/bin/bash", "bin/init.sh"],
   "customizations": {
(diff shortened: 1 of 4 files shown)


lalver1 (dismissed):
This looks good! 👍  I tested by rebuilding the `eligibility-server` images locally (the dev container ran without a problem) and then using the `eligibility_server` test image for my local benefits app. I didn't test the terraform change but I'm sure it works since it looks just like the terraform change in benefits ([PR 2156](https://github.com/cal-itp/benefits/pull/2156)).

The only thing I saw is that in the `compose` file of benefits we use `./:/calitp/app` for the volume mount and here we use `.:/calitp/app/`. They are both equivalent so it doesn't matter but if we wanted to make them the same we could.

thekaveman:
@lalver1:

> The only thing I saw is that in the compose file of benefits we use ./:/calitp/app for the volume mount and here we use .:/calitp/app/. They are both equivalent so it doesn't matter but if we wanted to make them the same we could.

Thanks for that feedback, I fixed the mounts to be the same as in `benefits` and rebased.

lalver1 (approved):
Looks great!

PR #464, opened 2024-06-13
Chore: Create an issue template for new releases

### LINKED ISSUE #433: Create an issue template for new releases
Based on https://github.com/cal-itp/benefits/blob/dev/.github/ISSUE_TEMPLATE/release.yml

Noting the difference here that we don't need a branch/version bump as that comes from the annotated git tag. Generally follows the process in these docs https://github.com/cal-itp/littlepay?tab=readme-ov-file#release-a-new-version

See a recent example in #434

lalver1:
Closes #433 

This PR creates an [issue form template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms) for new releases.

### DIFF
--- a/.github/ISSUE_TEMPLATE/release.yml
+++ b/.github/ISSUE_TEMPLATE/release.yml
@@ -1,0 +1,66 @@
+name: New Release
+description: Publish a new release of eligibility-server
+title: Make a Release
+labels:
+  - release
+body:
+  - type: markdown
+    attributes:
+      value: |
+        ## Prepare a new release
+
+        Use the form below to prepare a new release of eligibility-server.
+
+        Each release is coordinated by a **Release Manager**. The release manager may assign sub-tasks or ask for help
+        as-needed, but is otherwise responsible for all aspects of the release.
+
+        Each release also identifies a **Smoke Tester** responsible for carrying out Smoke Tests.
+
+        After this issue is created, use the checklist to manage the steps
+        of the release process, marking items as completed. [Read more about the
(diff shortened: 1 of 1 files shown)


lalver1:
One small thing to note is that this template's checklist contains an item for the release candidate and another for the regular release. This may cause confusion about what to use in the title of the ticket (`Make a Release: YYYY.0M.R` or `Make a Release: YYYY.0M.-rcR` ), but it should be clear that the ticket should have the regular release version in the title, not the release candidate version.

We could add `Release Type` to the form (similar to [benefits](https://github.com/cal-itp/benefits/blob/dev/.github/ISSUE_TEMPLATE/release.yml) but with `release-candidate` and `release` instead of `Regular` and `Hotfix`) but it seems too much to have to create two separate issues, one for release candidates and one for releases.

thekaveman:
> ...it should be clear that the ticket should have the regular release version in the title, not the release candidate version.
>
> ...it seems too much to have to create two separate issues, one for release candidates and one for releases.

Yeah we definitely don't want two separate issues! Since the checklist includes an item for making a release candidate, I think you're right and we're good with just the one template as-is.

thekaveman (commented):
@lalver1, can you see if it is possible to have a description / body in the issue that gets created?

E.g. it would be great if the text of the `markdown` field showed up in the issue, not just in the form for creating an issue from the template.

angela-tran (approved):
This looks great to me! It'll be so nice to use this rather than the copy-pasting we've been doing.

Issue #480, opened 2024-07-11
Infra pipeline is broken - deployment job requires an environment


angela-tran:
The infra pipeline is currently not able to run. The actual eligibility servers are still up and running, so this is not causing an outage, but we should get the infra pipeline back to working ASAP.

## To Reproduce

1. Go to Azure DevOps
2. View recent infra pipeline runs
3. See error about the job requiring an [Environment](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/environments?view=azure-devops#create-an-environment)

<img src="https://github.com/cal-itp/eligibility-server/assets/25497886/156c5a71-e6de-449e-b922-8a77ff1a2a25" width="400">

## Expected behavior

The infra pipeline succeeds.

## Environment

Affects the infra pipeline only

## Agency

Applies to infra pipelines for all agencies

## Additional context

I removed the `Approval` environment from the `Apply` deployment job in #477 in an attempt to have it run without needing our manual approval, but that is not how it works. 

[Environments](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/environments?view=azure-devops#create-an-environment) in Azure DevOps represent a collection of resources that you might want to deploy to, for example maybe a VM or a Kubernetes environment. The Environment does not actually need to hold any resources, but it can. A use case for having an empty Environment is so you can use other features that it provides like deployment history and manual approvals.

A [deployment job](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/deployment-jobs?view=azure-devops) is a special type of [job](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=azure-devops&tabs=yaml), and when it runs, it adds to the deployment history of the Environment that it is associated with. Therefore it requires an Environment.

**The simplest way to remove approvals**:
 - add back `environment: Approval` to the `Apply` job
 - edit the `Approval` environment to not require any approvals (screenshot of configuration screen below)
<img src="https://github.com/cal-itp/eligibility-server/assets/25497886/cb1301f0-0af1-4cf6-a95d-6ff5cca8bc6a" width="500">

**An alternate way** which requires more changes, but uses Environments in a way that maybe makes more sense:
 - create 3 new Environments: `Dev`, `Test`, and `Prod` - none of which require approval
 - in `terraform/pipeline/deploy.yml`, derive the correct Environment name and store it in a variable
 - set the `environment` on the `Apply` job using that variable
 - delete the `Approval` environment

[...]

PR #483, opened 2024-07-11
Refactor: use dev, test, and prod for deploy environment

angela-tran:
Part of #480

### DIFF
--- a/terraform/pipeline/deploy.yml
+++ b/terraform/pipeline/deploy.yml
@@ -24,6 +24,9 @@ stages:
               WORKSPACE=$(python terraform/pipeline/workspace.py)
               echo "##vso[task.setvariable variable=workspace;isOutput=true]$WORKSPACE"
 
+              ENVIRONMENT=$(case $WORKSPACE in "dev") echo "dev" ;; "test") echo "test" ;; "default") echo "prod" ;; *) echo "dev" ;; esac)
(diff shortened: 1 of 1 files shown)


terraform/pipeline/deploy.yml
```diff
@@ -99,10 +104,12 @@ stages:
     jobs:
       - deployment: Apply
         condition: succeeded()
-        environment: Approval
+        environment: $(environment)
```
angela-tran, line 107: Not sure if this will actually work. Seeing what the pipeline says when it runs this.

angela-tran: > Job Apply: Environment $(environment) could not be found. The environment does not exist or has not been authorized for use.

image

terraform/pipeline/deploy.yml
```diff
@@ -99,10 +104,12 @@ stages:
     jobs:
       - deployment: Apply
         condition: succeeded()
-        environment: Approval
+        environment: ${{ variables.environment }}
```
angela-tran, line 107: With this compile-time syntax, the variable hasn't been evaluated to its runtime value yet. (It's trying to use the value on lines 93-94 as a literal value.)

image

angela-tran:
https://github.com/cal-itp/eligibility-server/issues/480#issuecomment-2223980071

PR #508, opened 2024-12-18
Chore: more logging around user lookup logic

angela-tran:
Related to cal-itp/benefits#2598

### DIFF
--- a/eligibility_server/verify.py
+++ b/eligibility_server/verify.py
@@ -108,6 +108,8 @@ class Verify(Resource):
             sub = self.hash.hash_input(sub)
             name = self.hash.hash_input(name)
 
+            logger.debug(f"Hashed sub, name: {sub},{name}")
+
         existing_user = User.query.filter_by(sub=sub, name=name).first()
         if existing_user:
             existing_user_types = [type.name for type in existing_user.types]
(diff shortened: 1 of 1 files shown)


eligibility_server/verify.py
```diff
@@ -103,18 +103,24 @@ def _check_user(self, sub, name, types):
...
             name = self.hash.hash_input(name)
 
+            logger.debug(f"Hashed sub, name: {sub},{name}")
+
         existing_user = User.query.filter_by(sub=sub, name=name).first()
         if existing_user:
             existing_user_types = [type.name for type in existing_user.types]
+            logger.debug(f"Existing types on user: {existing_user_types}")
         else:
             existing_user_types = []
```
thekaveman, line 120: Maybe add one more here `User not found` or something

thekaveman (changes requested):

eligibility_server/verify.py
angela-tran: Added that log statement and another one to cover success case

https://github.com/cal-itp/eligibility-server/pull/508/commits/7c950fa4076d9e6aa08129866af76987cc6e624a

thekaveman (approved):
Good first step at least :+1: Hopefully we can learn some more

PR #512, opened 2024-12-19
Feat: metadata endpoint

### LINKED ISSUE #510: Store and expose the database load timestamp
It would be helpful to know the timestamp of when the database was last (re)loaded.

Idea:

* During `db-init`, create a new table to store a single record with the current timestamp when `db-init` is called
* Create an endpoint like `/metadata` that returns a JSON blob like:

```jsonc
{
  "app": {
    "version": "version string"
  },
  "db": {
    "timestamp": "stored timestamp",
    "users": 123, //count of users in DB
    "eligibility": ["list", "of", "types"]
   }
}
```

## Acceptance Criteria

- [ ] We have an easy way to tell when the database was last refreshed, and some basic stats about the current state of the app/database

## Additional context

Related to debugging cal-itp/benefits#2598

thekaveman:
Closes #510 

## Testing this PR

1. Checkout the branch locally, open in devcontainer
2. Reset the DB: `flask drop-db` then `flask init-db`
3. Confirm you see a record in the `metadata` table in the Sqlite database `code -r /tmp/test.db`
4. Launch the app with `F5`
5. Hit the `/metadata` endpoint, confirm output is as expected

### DIFF
--- a/eligibility_server/app.py
+++ b/eligibility_server/app.py
@@ -9,6 +9,7 @@ from flask.logging import default_handler
 from flask_restful import Api
 
 from eligibility_server import __version__, db, sentry
+from eligibility_server.db.models import Metadata
 from eligibility_server.keypair import get_server_public_key
 from eligibility_server.settings import Configuration
 from eligibility_server.verify import Verify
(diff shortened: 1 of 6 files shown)


thekaveman:
See https://github.com/cal-itp/eligibility-server/issues/510#issuecomment-2552513833 for why this is in Draft.

thekaveman:
Getting an error in the infra pipeline:

```log
Error: Error acquiring the state lock
│ 
│ Error message: state blob is already locked
│ Lock Info:
│   ID:        7af48bbd-8334-3356-becd-b6fe48e7a429
│   Path:      tfstate/terraform.tfstateenv:dev
│   Operation: OperationTypePlan
│   Who:       vsts@fv-az420-271
│   Version:   1.8.5
│   Created:   2024-12-20 20:34:46.015548148 +0000 UTC
```

It looks like during the key rotation last month, a state lock was made that was never released.

I'm going to [manually release the lock](https://developer.hashicorp.com/terraform/language/state/locking#force-unlock).

angela-tran (approved):
Tested this locally, and it looks good to me :+1:

PR #536, opened 2025-04-11
Chore: upgrade Terraform tasks

lalver1:
This PR upgrades Terraform task `TerraformTaskV3@3` which doesn't support Azure DevOps Workload identity federation to `TerraformTaskV4@4`which does support it.

## Notes
- According to this [Workload identity federation support table](https://learn.microsoft.com/en-us/azure/devops/pipelines/release/troubleshoot-workload-identity?view=azure-devops#review-pipeline-tasks), there is no need to update `AzureAppServiceManage@0` because it already supports identity federation. 
- `TerraformInstaller@0` does not seem to need identity permissions so an upgrade is not required.
- [`TerraformTaskV4@4`](https://github.com/microsoft/azure-pipelines-terraform/tree/main/Tasks/TerraformTask/TerraformTaskV4) still supports all the commands we use: `init`, `custom`, `plan`, `apply`. The parameters also seem to be the same.

### DIFF
--- a/terraform/pipeline/deploy.yml
+++ b/terraform/pipeline/deploy.yml
@@ -35,8 +35,8 @@ stages:
             displayName: Install Terraform
             inputs:
               terraformVersion: 1.8.5
-          # https://github.com/microsoft/azure-pipelines-terraform/tree/main/Tasks/TerraformTask/TerraformTaskV3#readme
-          - task: TerraformTaskV3@3
+          # https://github.com/microsoft/azure-pipelines-terraform/tree/main/Tasks/TerraformTask/TerraformTaskV4#readme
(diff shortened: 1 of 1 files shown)


terraform/pipeline/deploy.yml
```diff
@@ -36,7 +36,7 @@ stages:
             inputs:
               terraformVersion: 1.8.5
           # https://github.com/microsoft/azure-pipelines-terraform/tree/main/Tasks/TerraformTask/TerraformTaskV3#readme
```
thekaveman, line 38: We should update these comments as well please :pray:

thekaveman (changes requested):

thekaveman (approved):

PR #558, opened 2025-07-29
Docs: use docs color theme from Benefits

### LINKED ISSUE #539: Update documentation theme/colors
Match the recent updates to [`benefits`](https://github.com/cal-itp/benefits/pull/2595)

machikoyasuda:
closes #539 

## What this PR does
- Turns dark mode off
- Uses the Benefits docs' theme.css file

image

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -9,20 +9,7 @@ theme:
     - navigation.tabs
     - toc.integrate
   palette:
-    - media: "(prefers-color-scheme: light)"
-      scheme: default
-      primary: blue
-      accent: amber
-      toggle:
-        icon: material/toggle-switch-off-outline
-        name: Switch to dark mode
-    - media: "(prefers-color-scheme: dark)"
-      scheme: slate
-      primary: blue
-      accent: amber
-      toggle:
-        icon: material/toggle-switch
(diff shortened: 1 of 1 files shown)


mkdocs.yml
```diff
@@ -38,6 +25,7 @@ extra_javascript:
 
 extra_css:
   - https://use.fontawesome.com/releases/v5.13.0/css/all.css
+  - https://docs.calitp.org/benefits/styles/theme.css
```
machikoyasuda, line 28: @thekaveman @angela-tran @lalver1 

So we now have this Benefits docs theme CSS file, https://docs.calitp.org/benefits/styles/theme.css, which adds a few color changes to the existing MK Docs theme we are using. We want to use this theme across all Cal-ITP Benefits-related repos. One way to do it is to create the `theme.css` file in each of the docs directories, effectively copying and pasting styles into each docs repo. That felt un-DRY, so I wanted to try this way. But this also feels brittle, in case, for example, the URLs to Cal-ITP properties change or move. 

Do y'all have any other suggestions or ideas?

machikoyasuda: Moving forward with this above approach after discussing with @thekaveman

PR #474, opened 2024-07-09
fix(devcontainer): configure git to install pre-commit

machikoyasuda:
fixes #473 

I followed what @lalver1 did for Benefits here https://github.com/cal-itp/benefits/pull/2027 to fix this bug I got when I loaded the devcontainer on my macOS

PR #616, opened 2026-05-12
fix: move git dir command out of postAttach script and into Dockerfile

jgravois:
analogous to https://github.com/cal-itp/benefits/pull/3630

i'm under the impression that this is needed because >=git@2.35 tightened up security restrictions related to ambiguous directory ownership and `pip` makes calls to `git` under the hood.

ref: https://cal-itp.slack.com/archives/C037Y3UE71P/p1778616900243949?thread_ts=1778615995.226209&cid=C037Y3UE71P

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -9,5 +9,7 @@ COPY .git .git
 COPY eligibility_server/ eligibility_server/
 COPY pyproject.toml pyproject.toml
 
+RUN git config --global --add safe.directory /calitp/app
+
 # install devcontainer requirements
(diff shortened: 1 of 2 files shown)


angela-tran (approved):
> i'm under the impression that this is needed because >=git@2.35 tightened up security restrictions related to ambiguous directory ownership and pip makes calls to git under the hood.

Yeah, it's pretty opaque what changed exactly to cause this to show up earlier. I agree it's probably some internal git change.

Linking to https://github.com/cal-itp/eligibility-server/pull/474 too just for posterity

